PR #141, opened 2018-01-12
First working version of flattening converter

christophblessing:
I want to merge my version of the flattening converter so you can review my code. It does not produce a complete flattening but I would like to know if I am on the right track so far. I also added example files for testing.
Feedback is appreciated.


matthiaskoenig:
Hi @christophblessing,
great working on the flattening.

We recently encoded some DFBA models in SBML comp (with some of the submodels also using fbc).
https://www.biorxiv.org/content/early/2018/01/08/245076
Also some toy models included (see for instance Figure 1 of the paper).

So if you want to use the models as test models these are available from
https://github.com/matthiaskoenig/dfba/tree/master/models
as COMBINE archives (omex files in the models subfolders).

Feel free to use them for testing. Also some larger scale models included.
Best Matthias

skeating:
There are hundreds of comp files with their flat equivalents here

https://sourceforge.net/p/sbml/code/HEAD/tree/trunk/libsbml/src/sbml/packages/comp/util/test/test-data/

It would make sense to use the same test files as libsbml

Sarah

christophblessing:
Thanks, this is a great help!

PR #162, opened 2019-03-19
Fixed an incorrect error code in QualModelPluginConstraints.java

bhavyejain:
QUAL_20102 was not a related rule and was changed to QUAL_20101.
Avoids addition of unnecessary error codes.


niko-rodrigue:
Perfect, the changes are minimal, just what is needed and the commit message is fine.

PR #163, opened 2019-04-03
added implementation for MULTI_20201

bhavyejain:
Implementing logic for rule MULTI_20201 in ```MultiModelPluginConstraints```.
Reference taken from a similar implementation in ```MultiCompartmentPluginConstraints``` (Rule MULTI_20306)

### DIFF
--- a/extensions/multi/src/org/sbml/jsbml/validator/offline/constraints/MultiModelPluginConstraints.java
+++ b/extensions/multi/src/org/sbml/jsbml/validator/offline/constraints/MultiModelPluginConstraints.java
@@ -21,7 +21,6 @@ package org.sbml.jsbml.validator.offline.constraints;
 import java.util.Set;
 
 import org.sbml.jsbml.ListOf;
-import org.sbml.jsbml.SBMLDocument;
 import org.sbml.jsbml.ext.multi.MultiCompartmentPlugin;
 import org.sbml.jsbml.ext.multi.MultiConstants;
 import org.sbml.jsbml.ext.multi.MultiModelPlugin;
@@ -29,7 +28,6 @@ import org.sbml.jsbml.ext.multi.MultiSpeciesType;
 import org.sbml.jsbml.validator.SBMLValidator.CHECK_CATEGORY;
 import org.sbml.jsbml.validator.offline.ValidationContext;
 import org.sbml.jsbml.validator.offline.constraints.helper.DuplicatedElementValidationFunction;
-import org.sbml.jsbml.validator.offline.constraints.helper.UnknownAttributeValidationFunction;
 import org.sbml.jsbml.validator.offline.constraints.helper.UnknownElementValidationFunction;
 
 /**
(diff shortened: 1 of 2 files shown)


niko-rodrigue:
You implemented the constraint and it could work like that but I will make several comments inline of things that should be improved.

extensions/multi/src/org/sbml/jsbml/validator/offline/constraints/MultiModelPluginConstraints.java
```diff
@@ -90,10 +93,11 @@ public void addErrorCodesForAttribute(Set<Integer> set, int level,
       func = new UnknownAttributeValidationFunction<MultiModelPlugin>() {
```
niko-rodrigue: Here you have an UnknownAttributeValidationFunction which is not useful. You are inheriting from a specific ValidationFunction that is done to check unknown attribute. In this case, you do not need to create an anonymous class, you can directly use DuplicatedElementValidationFunction for the 'func' attribute. If I am not clear in my explanation, you can have a look at MultiSpeciesTypeConstraints, specifically MULTI_20401-3 and MULTI_20406 where we use directly a ValidationFunction implemented elsewhere and we just create a new instance.

extensions/multi/src/org/sbml/jsbml/validator/offline/constraints/MultiModelPluginConstraints.java
```diff
@@ -90,10 +93,11 @@ public void addErrorCodesForAttribute(Set<Integer> set, int level,
...
         @Override
         public boolean check(ValidationContext ctx, MultiModelPlugin multiM) {
-
-          // TODO
+        	
           // There may be at most one ListOfSpeciesTypes container object within a Model object
-          return true;
+          ValidationFunction<SBMLDocument> f = new DuplicatedElementValidationFunction<SBMLDocument>(MultiConstants.listOfSpeciesTypes);
+             
+          return f.check(ctx, multiM.getParent());
```
niko-rodrigue: Here for the check, you are using multiM.getParent() which is not correct. It could work like that but we want to make it work on multiM directly. Because the parent will be the Model and we want to store the data on the MultiModelPlugin instance.
To have the DuplicatedElementValidationFunction check to work (check the source code of this class), you need to modify the MultiParser class to add the XML elements as they are read into a list stored in the user objects that each SBase can have.

On https://github.com/sbmlteam/jsbml/blob/361d88121d8781709abd40b04f31006b0f77b336/extensions/multi/src/org/sbml/jsbml/xml/parsers/MultiParser.java#L180, you need to add

// keep order of elements for later validation
AbstractReaderWriter.storeElementsOrder(elementName, multiModel);

After that, the check should be nicely implemented and working fine.

extensions/multi/src/org/sbml/jsbml/validator/offline/constraints/MultiModelPluginConstraints.java
bhavyejain: I'll just have a look at these two rules for clarity.
The ```UnknownAttributeValidationFunction``` was written already (before my work) so I did not change it. 
To resolve this, I guess I just need to delete the anonymous function block and retain the  ```DuplicatedElementValidationFunction```?

extensions/multi/src/org/sbml/jsbml/validator/offline/constraints/MultiModelPluginConstraints.java
bhavyejain: I'll try to understand this. I'm having a bit of trouble getting this right now.

niko-rodrigue: You should check the implementation of AbstractReaderWriter.storeElementsOrder and DuplicatedElementValidationFunction.

bhavyejain: I went through the codes of both ```AbstractReaderWriter.storeElementsOrder``` and ```DuplicatedElementValidationFunction```, and as of what I understand, they are broadly serving the same purpose, of adding the elements to a list. 
The reason for adding elements to a list in ```MultiParser``` , according to me is that the parser fills the list with the elements relating to the specific extension and when ```DuplicatedElementValidationFunction``` is used, it gets the elements stored in the specific list of the model plugin of that extension.
Is this correct?

niko-rodrigue: Exactly. While we read the XML file, the elements are sent to the parser that will create the object instances and fill in the attributes and children. While the parser is doing that, we can store the elements in the order they are encountered which allow us to to the validation later on.

niko-rodrigue:
All seems good now.

PR #166, opened 2019-06-03
Finalize constraints for extended parameter class and extended species class.

bhavyejain:
I have added appropriate documentation at required places and made necessary changes in the existing code wherever required only.


extensions/spatial/src/org/sbml/jsbml/xml/parsers/SpatialParser.java
```diff
@@ -241,9 +244,15 @@ public boolean processAttribute(String elementName, String attributeName,
       }
       contextObject = spatialReaction;
     }
-
-    return super.processAttribute(elementName, attributeName, value, uri, prefix,
+    
+    isAttributeRead = super.processAttribute(elementName, attributeName, value, uri, prefix,
       isLastAttribute, contextObject);
+    
+    if(!isAttributeRead) {
```
niko-rodrigue: This should not need to be done, it is already done in https://github.com/sbmlteam/jsbml/blob/8ea32d6aef94624f0b9110e8a41a58303b2a8c99/core/src/org/sbml/jsbml/xml/stax/SBMLReader.java#L1196
So in those cases, the method processUnknownAttribute would be called twice!

extensions/spatial/src/org/sbml/jsbml/ext/spatial/SpatialSpeciesPlugin.java
```diff
@@ -174,12 +174,10 @@ public boolean readAttribute(String attributeName, String prefix, String value)
 
     if (attributeName.equals(SpatialConstants.isSpatial)) {
       try {
-        setSpatial(StringTools.parseSBMLBoolean(value));
+        setSpatial(StringTools.parseSBMLBooleanStrict(value));
         isAttributeRead = true;
-      } catch (Exception e) {
-        throw new SBMLException(
```
niko-rodrigue: You should still throw the SBMLException so that a warning message is printed to the user. You should also probably keep both catch clause, your IllegalArgumentException and as well the general Exception. In both cases, you need to throw as well the SBMLException with the warning message as it was.

niko-rodrigue:
You don't really need to close the request when you need to do some fixes. You can do more commits and they will be included in the push request.

bhavyejain:
I actually deleted the branch on my fork (which I should not have until this was merged) and hence it was closed.
I'll just correct it.

extensions/spatial/src/org/sbml/jsbml/xml/parsers/SpatialParser.java
bhavyejain: @niko-rodrigue  But somehow, this was not being done automatically. Only when I added this code, it registered the unknown attribute.

extensions/spatial/src/org/sbml/jsbml/ext/spatial/SpatialSpeciesPlugin.java
bhavyejain: Okay! I'll change this.

extensions/spatial/src/org/sbml/jsbml/xml/parsers/SpatialParser.java
niko-rodrigue: Now that your test is working, could you try to comment the call to super.processUnknownAttribute and check if the test is still working or not.

bhavyejain:
I rectified SpatialReactionPlugin.java and added throwing an exception, and catching general exceptions.
Please do have a look!

PR #167, opened 2019-06-03
implement extended reaction class constraints - 20601 & 20602

bhavyejain:
The test files for 20601 and 20602 had to be created manually. 20602 is working perfectly.
20601 is working perfectly for catching unknown attributes, but the call to ```isSetIsLocal()``` faced some issue while testing for ```spatial:isLocal``` to be present necessarily. The tests showed that the function was accessed and worked fine when ```isLocal``` is present, even with a wrong value, but was not accessed at all when the attribute was missing altogether. This might be a test case related problem since I hand made it. Please have a look at the code, because I'm certain it should work for a proper test case.

### DIFF
--- a/extensions/spatial/src/org/sbml/jsbml/ext/spatial/SpatialReactionPlugin.java
+++ b/extensions/spatial/src/org/sbml/jsbml/ext/spatial/SpatialReactionPlugin.java
@@ -28,6 +28,7 @@ import org.apache.log4j.Logger;
 import org.sbml.jsbml.PropertyUndefinedError;
 import org.sbml.jsbml.Reaction;
 import org.sbml.jsbml.util.StringTools;
+import org.sbml.jsbml.xml.parsers.AbstractReaderWriter;
 
 /**
  * @author Alex Thomas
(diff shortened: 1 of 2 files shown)


niko-rodrigue:
Can you attached the constraint files that you modified like this I can check them and add them to my set of test files so that the test work for me as well.

bhavyejain:
@niko-rodrigue I'll just confirm the test files and attach them here after the meeting.

bhavyejain:
@niko-rodrigue the link for new test files is:
https://drive.google.com/drive/folders/1HwZP-O1JSS-5JUPyDVrBo-33_saRIa52?usp=sharing

Could you please have a look at 1220601-fail-01-01 result? 
I am not able to figure out why the control is not even entering the `check()` when the spatial:isLocal attribute is missing! Putting a line `logger.warn("set local: "+srp.isSetIsLocal());` before the `if()` in `check()` chows that the control is not even reaching the if statement.
There does not seem to be a problem with the parser, because the second test file does not show a similar behaviour. It enters the check and also validates the unknown attribute.

bhavyejain:
https://github.com/sbmlteam/jsbml/blob/500be9dac9020b93a3848e67464bf06ada2bfbed/extensions/spatial/src/org/sbml/jsbml/ext/spatial/SpatialReactionPlugin.java#L77-L79

The constructor Boolean(boolean) is deprecated too. I hope this is not problematic?

PR #174, opened 2019-06-17
Implement rules 208XX and 218XX

bhavyejain:
Implemented:
DomainConstraints
AnalyticGeometryConstraints

Modified tests for rule spatial-20806 have been added to the head folder.

### DIFF
--- a/extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/DomainConstraints.java
+++ b/extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/DomainConstraints.java
@@ -1,0 +1,252 @@
+/*
+ * ----------------------------------------------------------------------------
+ * This file is part of JSBML. Please visit <http://sbml.org/Software/JSBML>
+ * for the latest version of JSBML and more information about SBML.
+ * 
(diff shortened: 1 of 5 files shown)


extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/DomainConstraints.java
```diff
@@ -0,0 +1,249 @@
...
+		  		// Apart from the general notes and annotations subobjects permitted on all SBML objects,
+		  		// a ListOfInteriorPoints container object may only contain InteriorPoint objects.
+		  		
+		  		func = new ValidationFunction<Domain>() {
+
+		  		    @Override
+		  		    public boolean check(ValidationContext ctx, Domain dom) {
+		  		    	
+		  		        if (dom.isSetListOfInteriorPoints()) {		  		        	
+		  		            boolean coreElements = new UnknownCoreElementValidationFunction<ListOf<InteriorPoint>>().check(ctx, dom.getListOfInteriorPoints());
```
niko-rodrigue: for both 20807 and 20808, you should be able to use the generic Unknown methods to have only one method call instead of two.

PR #175, opened 2019-06-18
Implement spatial-237XX

bhavyejain:
Implement GeometryConstraints.
Made a suggested change in DomainConstraints.

Modified test files 23706-23717 in the head folder.
Problem was that in folders 23706-23711, each folder contained test files for all of these rules,
and not the specific rule. Same for 23712-23717.

### DIFF
--- a/extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
+++ b/extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
@@ -34,6 +34,7 @@ import org.sbml.jsbml.validator.SBMLValidator.CHECK_CATEGORY;
 import org.sbml.jsbml.validator.offline.ValidationContext;
 import org.sbml.jsbml.validator.offline.constraints.helper.DuplicatedElementValidationFunction;
 import org.sbml.jsbml.validator.offline.constraints.helper.InvalidAttributeValidationFunction;
+import org.sbml.jsbml.validator.offline.constraints.helper.UnknownAttributeValidationFunction;
 import org.sbml.jsbml.validator.offline.constraints.helper.UnknownCoreAttributeValidationFunction;
 import org.sbml.jsbml.validator.offline.constraints.helper.UnknownCoreElementValidationFunction;
 import org.sbml.jsbml.validator.offline.constraints.helper.UnknownElementValidationFunction;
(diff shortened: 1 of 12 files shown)


bhavyejain:
@niko-rodrigue  Sorry for that last commit. I had not pulled the correction to my local branch (only to local master). So I made the change and committed it and it gave a merge conflict. I retained the change that you committed to my repo and was already merged in the master here.

extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
```diff
@@ -0,0 +1,420 @@
...
+		  		func = new ValidationFunction<Geometry>() {
+		  			
+		  			@Override
+		  			public boolean check(ValidationContext ctx, Geometry geom) {
+		  				
+		  				boolean locc = geom.isSetListOfCoordinateComponents() 
+		  						&& new DuplicatedElementValidationFunction<Geometry>(SpatialConstants.listOfCoordinateComponents).check(ctx, geom);
+		  				boolean lodt = geom.isSetListOfDomainTypes() 
+		  						&& new DuplicatedElementValidationFunction<Geometry>(SpatialConstants.listOfDomainTypes).check(ctx, geom);
+		  				boolean lod = geom.isSetListOfDomains()
```
niko-rodrigue: I don't think this logic is right. If a ListOf is not set, the boolean should be true and in your case, it is false.
You can create an additional test file where one or several of the listOf are not there at all and you will see that your current code will fail the constraint where it should not.

extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
```diff
@@ -0,0 +1,420 @@
...
+		  		// metaid and sboTerm. No other attributes from the SBML Level 3 Core namespaces are permitted 
+		  		// on a ListOfCoordinateComponents object.
+		  		
+		  		func = new ValidationFunction<Geometry>() {
+		  			
+		  			@Override
+		  			public boolean check(ValidationContext ctx, Geometry geom) {
+		  				
+		  				if(geom.isSetListOfCoordinateComponents()) {
+		  					return new UnknownCoreAttributeValidationFunction<ListOf<CoordinateComponent>>().check(ctx, geom.getListOfCoordinateComponents())
```
niko-rodrigue: You can use the generic method there, no need to call both methods. Same for the following constraints.

extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
bhavyejain: @niko-rodrigue I just checked in the specification, and it says that listOfCoordinateComponents is always needed and rest are optional. So I think I'll retain the isSetListOf method for coordinate component and remove the method for rest of them?

extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
bhavyejain: Okay, sure!

extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
```diff
@@ -0,0 +1,410 @@
...
+		  		// ListOfDomainTypes, ListOfDomains, ListOfAdjacentDomains, ListOfGeometryDefinitions and
+		  		// ListOfSampledFields elements. No other elements from the SBML Level 3 Spatial Processes
+		  		// namespaces are permitted on a Geometry object.
+		  		
+		  		func = new ValidationFunction<Geometry>() {
+		  			
+		  			@Override
+		  			public boolean check(ValidationContext ctx, Geometry geom) {
+		  				
+		  				boolean locc = geom.isSetListOfCoordinateComponents()
```
niko-rodrigue: You still need to have a `if (isSetListOf)` for each test because otherwise it does create an empty list that you don't want to create and go through a method call that is not necessary.

If the specific rule 23704 does not ask you to test if listOfCoordinateComponents is present, just if it is not duplicated, you should not test more. Don't fail a check for reason that are not part of the rule :-)

niko-rodrigue: You could have one boolean only defined to true at the start, then you use it for each if statement:
`
boolean check = true;
if (check && isSetListof)  {
check &= "duplicated test"
}
...

return check;
`

bhavyejain: Oh, right! I understand it now! :-)
That check for listOfCoordinateComponents can be added later as a rule added by hand.
We are checking for duplicates only if that ListOf<> element is set, right?

bhavyejain:
@niko-rodrigue changes done!

extensions/spatial/src/org/sbml/jsbml/validator/offline/constraints/GeometryConstraints.java
niko-rodrigue: yes

PR #189, opened 2019-12-09
Vetter comp single file

niko-rodrigue:


niko-rodrigue:
Some failing tests for me so waiting to resolve those before merging.

PR #191, opened 2020-01-08
Pivotal 161592475

thortiede:
Change the handling of lowerBound and upperBound deletion during conversion from v2 to v1 to be deleted after all reactions have been processed.


niko-rodrigue:
Looking good.

PR #192, opened 2020-01-09
Offline validator meta id bugfix

Onur101010:
A bug in the initialization of patterns for the meta Id resulting in SBML Documents beeing evaluated as invalid was solved.


Onur101010:
The `System.out.`-prints in the `initMetaIdPatterns()` method are for the reviewer to better understand what is going on in the testclass and can be deleted after the review.

niko-rodrigue:
@Onur101010 the change seem to fix the problem so that's good.
Although, I see some problems that need to be fixed with the current code:

-  You are not covering the case of SBML L2V1, there was a metaid already and it should follow the rules of L2V2.
- If the simple pattern does not change, you should initialize it only once.
- With the current code, if you validate many models with different level and version, you will initialize the pattern many times and if you are in a web server that could do validations in parallel it can become problematic. You basically have two cases SBML L2V1 and L2V2 and then anything above L2V2 so you could have two variables metaIdPatternL2V2 and metaIdPatternL2V3 or you could have a hashmap with a pair a level and version as key and the pattern as value. You should just make sure that you initialize the metaid patterns only twice even if thousand of validations are performed.

Onur101010:
There is a hashmap now with two possible keys "L2V1" which are for L2V1, L2V2 patterns and "L2V3" which is for all patterns above L2V3. 

The `initMetaIdPatterns` method now returns the key which can be used to detect which pattern was loaded. This is used in `isValidMetaId` to match the correct patterns.

core/src/org/sbml/jsbml/validator/SyntaxChecker.java
```diff
@@ -204,31 +214,61 @@ public static final boolean isValidId(String idCandidate, int level,
...
 
-    // In the most cases the first check will be sufficient.
-    if (syntaxChecker.simpleMetaIdPattern.matcher(idCandidate).matches()) {
-      return true;
-    }
-    /* In the worst case, we have to perform two checks, but this is
-     * hopefully much faster than doing the full check each time.
-     * It can be assumed that the majority of allowable characters is
-     * not used within these identifiers.
+    String metaId = syntaxChecker.initMetaIdPatterns(level, version);
```
niko-rodrigue: @Onur101010  calling this string metaId is quite misleading. It is always better to try to give a name describing the variable properly. In this case, I would call it something like metaIdPatternKey.

core/test/org/sbml/jsbml/xml/test/MetaIdValidationTest.java
```diff
@@ -0,0 +1,129 @@
...
+   * u0967 is a devanagari digit of type Digit 
+   */
+  
+  @Test
+  public void initMetaPatternsLevelVersionTest() {
+    int level, version;
+
+    //specifications L2V2 under 3.1.6
+    //has to start with a letter or an underscore
+
```
niko-rodrigue: Please add a test for L2V1 as well here.

core/src/org/sbml/jsbml/validator/SyntaxChecker.java
```diff
@@ -308,39 +348,147 @@ private void initL1V2SNamePattern() {
...
+        if(!syntaxChecker.simpleMetaIdPatterns.containsKey("L2V1")) {
+
+          simpleNameChar = "[" + letter + digit + dot + dash + underscore + "]";
+          simpleMetaId = "[" + letter + underscore + "]" + "[" + simpleNameChar + "]*"; 
+          simpleMetaIdPatterns.put("L2V1", Pattern.compile(simpleMetaId)) ;  
+        }
+
+        return "L2V1";
+      }
+      System.out.println("already existing Pattern L2V1, do nothing");
```
niko-rodrigue: Be careful to remove the System.out when you are ready to merge

niko-rodrigue:
Last small bit would be to remove the System.out in the MetaIdValidationTest class. And to add the test in the core test suite (I can do that later as well).

PR #194, opened 2020-01-29
Render font family

Onur101010:
Changed `FontFamily `enum to support the Strings "monospace", "sans-serif" and "serif" instead of the uppercase names of the enum values. 
Also changing `FontFamily` values to String is easier now because of the extra stored value (String), simply using `toString()` is enough.  

In addition to that a new method `toFontFamily()` matches Strings and gives the value back as `FontFamily` if it is contained in the enum. 

The standard storing value is String, therefore all classes using `FontFamily` as standard were updated too.


niko-rodrigue:
The changes look fine. One thing that you should check is that you write the proper string to XML, like defined in the specifications and that you can read it again from the file.
You could add another test that read from a file and write it to a String and you can read the SBMLDocument again from this String to check that the font family is still there.

Onur101010:
I tried setting up a dummy `SBMLDocument` with a `font-family` and then writing it to `String` and back. 
Seems to work but when using `writeXMLAttributes()` of the `Text` Class to set up the attributes problems occur in relation to the prefix of the keys.

So I tried to do it manually without prefixes in the keys, this way it works.

Onur101010:
I added a testclass with a dummy document with the font-family attribute set. The second test `writeAndReadTest()` first writes the document to a `String` and then reads it back and afterwards tests if the `font-family` property is set correctly.

PR #200, opened 2020-03-11
Check render

DEVetter:
For pivotal-stories #135708429 and #153405156


DEVetter:
The question about Event-forwarding (see Ellipse-class) should be cleared before finally merging this.

PR #203, opened 2020-03-17
Improved layout package and example-implementations for render.director

DEVetter:
a) Improved layout by adding unset-methods etc to complete API
b) Added example-implementations (now with zbit-colors in the render-example)
c) Fixed RelAbsVector-bug


extensions/layout/src/org/sbml/jsbml/ext/layout/BoundingBox.java
```diff
@@ -364,7 +381,7 @@ public void setPosition(Point point) {
       logger.warn(MessageFormat.format(
         ResourceManager.getBundle("org.sbml.jsbml.resources.cfg.Messages").getString("UNDEFINED_ATTRIBUTE"),
         "name", getLevel(), getVersion(), getElementName()));
-      // TODO: This must be generally solved. Here we have an SBase with ID but without name!
+      // TODO 2013/12: This must be generally solved. Here we have an SBase with ID but without name!
```
niko-rodrigue: Not sure why 2013/12 was added here. But in general, this is probably something that need to be reviewed as now in L3V2 everything can have id and name. And may be the layout package changed in the final specification as well.

extensions/layout/src/org/sbml/jsbml/ext/layout/CubicBezier.java
```diff
@@ -278,8 +284,12 @@ public Point removeBasePoint1() {
   }
 
   /**
+   * Removes the basepoint 2 from this {@link CubicBezier} 
+   * <br/>
```
niko-rodrigue: Instead of using <br/>, try to just let an empty line and then put the next section of code between <p>...</p>
I think I had problem in maven with <br/>.

extensions/layout/src/org/sbml/jsbml/ext/layout/Dimensions.java
```diff
@@ -231,21 +232,21 @@ public boolean isIdMandatory() {
    * @return
    */
   public boolean isSetDepth() {
-    return !Double.isNaN(depth);
+    return depth != null && !Double.isNaN(depth);
```
niko-rodrigue: Not sure we should test `NaN` here. As it can be value set in purpose if the number is not known.

extensions/layout/src/org/sbml/jsbml/ext/layout/GeneralGlyph.java
```diff
@@ -710,6 +711,7 @@ public void setListOfReferenceGlyph(ListOf<ReferenceGlyph> listOfReferenceGlyph)
   /**
    * Unsets the {@link #listOfReferenceGlyphs}.
    */
+  // TODO 2020/03: why is this private?
```
niko-rodrigue: check with libsbml if they have this method public. It might just be a typo.

extensions/layout/src/org/sbml/jsbml/ext/layout/Layout.java
```diff
@@ -502,11 +502,11 @@ public boolean equals(Object object) {
    */
   @SuppressWarnings("unchecked")
   private <T> List<T> findGlyphs(ListOf<? extends T> listOfGlyphs, String id) {
-    // TODO - use the Model findUniqueNamedSBase method if model is not null ?
+    // TODO 2014/05: use the Model findUniqueNamedSBase method if model is not null ?
     if (isSetListOfReactionGlyphs() && (listOfGlyphs != null) && (!listOfGlyphs.isEmpty())) {
```
niko-rodrigue: You need to remove the `isSetListOfReactionGlyphs()` test as well. You can remove the TODO as well. 

The javadoc should be written as well to explain that what we return are glyphs that reference the given id. We are not looking for a glyph id here which might be confusing.

extensions/layout/src/org/sbml/jsbml/ext/layout/Point.java
```diff
@@ -296,7 +305,7 @@ public boolean isSetX() {
    * @return {@code true} if the y coordinate is set.
    */
   public boolean isSetY() {
-    return !Double.isNaN(y);
+    return y != null && !Double.isNaN(y);
```
niko-rodrigue: Same remark as previously. we should not test for NaN here.

PR #205, opened 2020-03-27
Render validation

DEVetter:
Implemented validation rules render-20201:render-23043 (i.e. missing the early rules render-10101:render-20103, also missing the recommendations render-21006/21007).

On the way, did some refactoring to the render-package. Some key refactorings:
 - dealing with incorrect attributes when reading: Added non-object-oriented helper-method `static void addToInvalidXMLUserObject` to XMLTools to register invalid attribute-values for later validation; and used it in the `readAttribute`-method
 - polygon's getChildAt can now handle the usual calls for notes/annotation (and is consistent with Polygon.getChildCount) -- this may need be translated to RenderCurve too, but RenderCurve did not cause problems here
 - Added UserObject-tracking in RenderParser, but for now just for RenderInformationBase (see respective TODO) -- this may need be added for the other Classes where subobjects are to be expected (cf. LayoutL3Parser)
 - Added `boolean isListOf___Empty()`-methods to complement the `boolean isSetListOf___()`-methods: `isListOf___Empty` will return `true` iff the list is not null, but empty (which is usually forbidden by the validation-rules). Thereby, the API as stood before does not change (`isSetListOf___` still behave the same).

The sbml-test-files were adapted where they (incorrectly?) made use of <g ... /> instead of <render:g ... />.

### DIFF
--- a/extensions/render/src/org/sbml/jsbml/ext/render/XMLTools.java
+++ b/extensions/render/src/org/sbml/jsbml/ext/render/XMLTools.java
@@ -23,7 +23,13 @@ package org.sbml.jsbml.ext.render;
 import java.awt.Color;
 import java.util.Locale;
 
+import org.sbml.jsbml.JSBML;
 import org.sbml.jsbml.util.StringTools;
+import org.sbml.jsbml.util.TreeNodeWithChangeSupport;
+import org.sbml.jsbml.xml.XMLAttributes;
+import org.sbml.jsbml.xml.XMLNamespaces;
+import org.sbml.jsbml.xml.XMLNode;
+import org.sbml.jsbml.xml.XMLTriple;
 
 /**
  * Utility class to help write the XML
@@ -193,6 +199,22 @@ public class XMLTools {
     }
     return temp;
   }
+  
+  
+  /**
+   * Checks whether given value is a valid doubleArray (i.e. a
+   * comma-and-space-separated list of doubles)
+   * 
+   * @param value
+   * @return
+   */
+  public static boolean canDecodeStringToArrayDouble(String value) {
+    String[] array = value.split(", ");
+    for(String s : array) {
+      try { Double.parseDouble(s); } catch (NumberFormatException e) { return false; }
+    }
+    return true;
+  }
 
   /**
    * 
@@ -223,6 +245,33 @@ public class XMLTools {
       temp[i] = StringTools.parseSBMLShort(array[i]);
     }
     return temp;
+  }
+  
+  public static boolean canDecodeStringToArrayUnsignedInt(String value) {
+    String[] array = value.split(", ");
+    for (String s : array) {
+      try { Integer.parseUnsignedInt(s); } catch(NumberFormatException e) { return false; };
+    }
+    return true;
+  }
+  
+
+  /**
+   * (not quite object-oriented style)
+   * Helper to add an entry to the {@link JSBML.INVALID_XML}-{@link XMLNode} on given TreeNode
+   * @param t to which to add the UserObject/where to modify the userObject
+   * @param attributeName
+   * @param invalidValue
+   */
+  public static void addToInvalidXMLUserObject(TreeNodeWithChangeSupport t, String attributeName, String invalidValue) {
+    if(!t.isSetUserObjects() || !t.containsUserObjectKey(JSBML.INVALID_XML)) {
+      // TODO 2020/03: this is somewhat abusive code
+      t.putUserObject(JSBML.INVALID_XML,
+        new XMLNode(new XMLTriple(JSBML.INVALID_XML, "", ""),
+          (XMLAttributes) null, (XMLNamespaces) null, 0l, 0l));
+    }
+    XMLNode invalidNode = (XMLNode) t.getUserObject(JSBML.INVALID_XML);
+    invalidNode.addAttr(attributeName, invalidValue);
   }
 
 }
--- a/extensions/render/src/org/sbml/jsbml/validator/offline/constraints/LinearGradientConstraints.java
+++ b/extensions/render/src/org/sbml/jsbml/validator/offline/constraints/LinearGradientConstraints.java
@@ -1,0 +1,131 @@
+/*
+ * ----------------------------------------------------------------------------
+ * This file is part of JSBML. Please visit <http://sbml.org/Software/JSBML>
+ * for the latest version of JSBML and more information about SBML.
+ *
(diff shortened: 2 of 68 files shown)


extensions/render/src/org/sbml/jsbml/validator/offline/constraints/TransformationConstraints.java
```diff
@@ -0,0 +1,87 @@
...
+      break;
+    case RENDER_22504:
+      func = new InvalidAttributeValidationFunction<Transformation>(RenderConstants.transform);
+      break;
+    case RENDER_22505:
+      func = new ValidationFunction<Transformation>() {
+        @Override // Any string
+        public boolean check(ValidationContext ctx, Transformation t) { return true; }
+      };
+      break;
```
draeger: Possibly forgotten `default: break;` here?

extensions/render/src/org/sbml/jsbml/validator/offline/constraints/LocalStyleConstraints.java
```diff
@@ -0,0 +1,86 @@
...
+      break;
+    case RENDER_21604:
+      func = new ValidationFunction<LocalStyle>() {
+        @Override
+        public boolean check(ValidationContext ctx, LocalStyle t) {
+          // Any string
+          return true;
+        }
+      };
+      break;
```
draeger: Do we need a `default: break;` here?

extensions/render/src/org/sbml/jsbml/validator/offline/constraints/PolygonConstraints.java
```diff
@@ -0,0 +1,94 @@
...
+      break;
+    case RENDER_21704:
+      func = new UnknownPackageElementValidationFunction<Polygon>(LayoutConstants.shortLabel) {
+        public boolean check(ValidationContext ctx, Polygon poly) {
+          return super.check(ctx, poly)
+            && new DuplicatedElementValidationFunction<Polygon>(
+              LayoutConstants.listOfCurveSegments).check(ctx, poly);
+        }
+      };
+      break;
```
draeger: `default: break;`?

extensions/render/src/org/sbml/jsbml/validator/offline/constraints/LinearGradientConstraints.java
```diff
@@ -0,0 +1,131 @@
...
+      break;
+    case RENDER_21409:
+      func = new ValidationFunction<LinearGradient>() {
+        @Override
+        public boolean check(ValidationContext ctx, LinearGradient gradient) {
+          return !gradient.isSetZ2() || gradient.getZ2().isSetAbsoluteValue()
+            || gradient.getZ2().isSetRelativeValue();
+        }
+      };
+      break;
```
draeger: `default: break;`?

extensions/render/src/org/sbml/jsbml/validator/offline/constraints/LocalRenderInformationConstraints.java
```diff
@@ -0,0 +1,125 @@
...
+        @Override
+        public boolean check(ValidationContext ctx, LocalRenderInformation lri) {
+          if (lri.isSetListOfLocalStyles()) {
+            return new UnknownCoreAttributeValidationFunction<ListOf<LocalStyle>>().check(
+              ctx, lri.getListOfLocalStyles());
+          }
+          return true;
+        }
+      };
+      break;
```
draeger: `default: break;`?

extensions/render/src/org/sbml/jsbml/validator/offline/constraints/RenderInformationBaseConstraints.java
```diff
@@ -0,0 +1,230 @@
...
+        @Override
+        public boolean check(ValidationContext ctx, RenderInformationBase t) {
+          if(t.isSetListOfLineEndings()) {
+            return new UnknownCoreAttributeValidationFunction<ListOf<LineEnding>>().check(ctx,
+                t.getListOfLineEndings());
+          }
+          return true;
+        }
+      };
+      break;
```
draeger: `default: break;`?

extensions/render/src/org/sbml/jsbml/validator/offline/constraints/StyleConstraints.java
```diff
@@ -61,11 +70,34 @@ public void addErrorCodesForAttribute(Set<Integer> set, int level,
...
+      };
+      break;
+    case RENDER_22805:
+    case RENDER_22806:
+    case RENDER_22807:
+      func = new ValidationFunction<Style>() {
+        @Override // any string
+        public boolean check(ValidationContext ctx, Style t) { return true; }
+      };    
+      break;
```
draeger: `default: break;`?

draeger:
Don't we need `default: break;` statements for all `switch` statements?

DEVetter:
> Don't we need `default: break;` statements for all `switch` statements?
We did not, but now we do. I added them and moved the initialisation `func=null` to the default statement.

extensions/render/src/org/sbml/jsbml/ext/render/XMLTools.java
```diff
@@ -224,5 +246,32 @@ public static String encodeArrayShortToString(Short[] array) {
...
+  /**
+   * (not quite object-oriented style)
+   * Helper to add an entry to the {@link JSBML.INVALID_XML}-{@link XMLNode} on given TreeNode
+   * @param t to which to add the UserObject/where to modify the userObject
+   * @param attributeName
+   * @param invalidValue
+   */
+  public static void addToInvalidXMLUserObject(TreeNodeWithChangeSupport t, String attributeName, String invalidValue) {
+    if(!t.isSetUserObjects() || !t.containsUserObjectKey(JSBML.INVALID_XML)) {
+      // TODO 2020/03: this is somewhat abusive code
```
niko-rodrigue: @DEVetter Why abusive code ? 
In fact, there was already a method to do that, in org.sbml.jsbml.xml.parsers.AbstractReaderWriter.processInvalidAttribute(String, String, String, String, Object).
You should use this method instead of your method in XMLTools. You could call .AbstractReaderWriter.processInvalidAttribute from XMLTools if you think it would be easier to find (but this should be in the user/developer guide in fact).

niko-rodrigue: Forget about that, I have change the way it will work slightly and I will make the changes myself.

PR #206, opened 2020-04-02
story #154807739: make clear if we encounter package that has partial or no validation support

Onur101010:


core/src/org/sbml/jsbml/validator/offline/constraints/SBMLDocumentConstraints.java
```diff
@@ -47,6 +48,8 @@
    * 
    */
   public static final transient String XML_DECLARED_ENCODING = "jsbml.validator.xml.declared.encoding";
+  
+  private HashMap<String, Integer> packageValidationAvailability;
```
draeger, line 54: Even though it is private, I recommend to always declare variables with the least specific type, which is in this case `Map`.

niko-rodrigue:
Have you tested it with several xml files to see if the proper error is thrown and if it is not thrown when it should not. You could create two folders that we will add into the set of validations files, name it like the others files with some fail and pass files

Onur101010:
> t. You could create two folders that we will add into the set of validations files, name it like the others files with some fail and pass file

No not yet, okay I'll start and keep you updated.

Onur101010:
Is it sufficient that the first activated package of the Document with no validation leads the constraint to be broken? Or do we want to have all packages to have no support to break the constraint 70001?

Onur101010:
Okay perfect, then you can maybe doublecheck the validator with the new folders, it seems to work now.

niko-rodrigue:
Yes, it is working but I would like, as  an exercise that you put the two
ValidationFunctions into one to see if you understood correctly how it is
working and to show later on an example of a ValidationFunction that handle
several validation errors at the same time.

Onur101010:
Is something like this what you mean ?

niko-rodrigue:
You are almost there. Here, you are re-using the same ValidationFunction twice, which is already good. What I want is to have only on ValidationFunction, called only once (meaning, you can remove the 'case CORE_70002:'). 
And then, in the implementation of the validationFunction, you are not allowed to test for the error code or to use CORE_70001 or CORE_70002 !

niko-rodrigue:
Perfect, see how it make the method much simpler.

Last small details, please add a small comment(s) to explain that you are doing both errors in the same method. Also, you now can remove CORE_70002 from addErrorCodesForCheck.

PR #215, opened 2020-05-24
quickly check if an SBase has a certain extension

Onur101010:


core/src/org/sbml/jsbml/AbstractSBase.java
```diff
@@ -1835,6 +1835,15 @@ public int hashCode() {
     return hashCode + prime * getLevelAndVersion().hashCode();
   }
 
+  /*
+   * (non-Javadoc)
+   * @see org.sbml.jsbml.SBase#hasExtension(java.lang.String)
+   */
+  @Override
+  public boolean hasExtension(String extName) {
+    return getModel().getExtension(extName) != null;
```
draeger, line 1844: Sometimes, an extension can be directly attached to the object itself, such as the extended species in FBC that has additional attributes. My suggestion would be that the method on `AbstractSBase` only checks if `this` object has any extension.

core/src/org/sbml/jsbml/SBase.java
```diff
@@ -697,6 +697,11 @@
    */
   @Override
   public int hashCode();
+  
+  /**
+   * Returns {@code true} if the given extension exists in the model
```
draeger, line 702: See my other comment, the extension can exist directly regarding the current object.

PR #224, opened 2021-06-18
Fixed issue with unit derivation

eikept:
**Issue:**

The issue was that units were not always derived correctly as there were issues in the `simplify()` and `merge()` methods  in the classes `UnitDefinition` and  `Unit` respectively. 

**Main Changes:**

- Methods `simplify()` and `merge()`  were rewritten as to fix aforementioned issue with the unit derivation where this implementation was largely based on the corresponding methods in LibSBML (C++).
- As there are structural differences between the `simplify()` and `merge()` methods in JSBML and LibSBML the methods `multiplyWith()` and `divideBy()` in `UnitDefinition` had to be changed appropriately.
- `testUnitSimplification()` was modified to test more extensively and  new test method `testUnitMerging()` was added to test the `merge()` method individually. 

**Other Changes:**

- The `root()` method from `Maths` was extended as to return just the radiant if the root exponent is 1 and to use the square/cubic root methods from the regular java `Math` class for root exponents of 2/3 respectively instead of always using `pow()` from `Math`. This leads to better results should the root exponent be 2 or 3. Additionally, where applicable, calls of `Math.pow()` were replaced with `Maths.root()` at various places in the code.

**Remaining Issues:**

Because of inaccuracies in the calculation with floating point numbers some multipliers will be slightly incorrect when the unit is derived.

### DIFF
--- a/core/src/org/sbml/jsbml/Unit.java
+++ b/core/src/org/sbml/jsbml/Unit.java
@@ -1041,8 +1041,8 @@ public class Unit extends AbstractSBase implements UniqueSId {
 
   /**
    * Merges two {@link Unit} objects with the same 'kind' attribute value into a
-   * single {@link Unit}.
-   * 
+   * single {@link Unit}. Java Implementation of the corresponding C++ method in LibSBML
+   *
    * @param unit1
    *            the first {@link Unit} object; the result of the operation is left as
    *            a new version of this {@link Unit}, modified in-place.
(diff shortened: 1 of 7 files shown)


core/src/org/sbml/jsbml/util/Maths.java
```diff
@@ -296,10 +296,21 @@ public static final double log(double number, double base) {
...
+      return radiant;
+    }
+    else if(rootExponent == 2d) {
+      return Math.sqrt(radiant);
+    }
+    else if(rootExponent == 3d) {
+      return Math.cbrt(radiant);
+    }
+    else {
+      return Math.pow(radiant, 1d / rootExponent);
```
draeger, line 312: The first `else` is not needed because the `if` case is only checked if no exception has been thrown. Also, the last `else` can be removed because it is only reached if no `return` statement has been executed.

draeger (changes requested):

core/src/org/sbml/jsbml/Unit.java
```diff
@@ -1041,182 +1041,101 @@ public static boolean isValidUnit(Model model, String unit) {
...
-        "Cannot merge units with different kind properties {0} and {1}. Units can only be merged if both have the same kind attribute or if one of them is dimensionless.",
-        k1, k2));
+
+      return;
+    }
+
+    // If units are not invalid but not of same kind don't merge
+    if(!unit1.getKind().equals(unit2.getKind())) {
+      return;
+    }
```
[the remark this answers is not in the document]
niko-rodrigue: An aditional comment on that. Not sure if you are planning to work on validation as well but you might want to make sure that unit validation is not impacted by changes in behavior. It would be good if not already done that @eikept has the offline validation checks setup so that he can see if unit validation is affected and if validation need to be ajusted after the changes.

core/src/org/sbml/jsbml/util/Maths.java
```diff
@@ -296,10 +296,21 @@ public static final double log(double number, double base) {
    * @return the rootExponent-th root of the radiant
    */
   public static final double root(double radiant, double rootExponent) {
-    if (rootExponent != 0) {
-      return Math.pow(radiant, 1 / rootExponent);
+    if (rootExponent == 0d) {
+      throw new ArithmeticException("Root exponent must not be zero.");
+    }
+    else if(rootExponent == 1d) {
```
matthiaskoenig, line 302: I don't like this large additional complexity to introduce a micro-optimization. 
This is basically the fallacy of premature optimization in action. "premature optimization is the root of all evil."
Did anybody complain this being too slow? There are other places to start. No optimization without benchmarking. I would revert this to the simple readible one-liner on the left. 
Probably it is even slower then before because the function cannot be inlined any more ;/

matthiaskoenig (commented):
Looks good as I understand the merging algorithm. Two minor issues I had
- did not like the `root` optimization
- did not understand the merging behavior with invalid units

core/src/org/sbml/jsbml/util/Maths.java
draeger: We should possibly benchmark it using SBSCL.

eikept: @matthiaskoenig The idea of this change was not improve the runtime but to improve the accuracy of the results as `Math.pow()` doesn't seem to use `Math.cbrt()` internally if called with 1/3 as the exponent. Example: If i wanted to calculate what the third root of 64 is `pow(64, 1d/3d)` would return 3.9999999999999996 while `cbrt(64)` would return 4.0 . Maybe that wasn't quite clear from my description. Nevertheless we could of course still benchmark it if desired.

core/src/org/sbml/jsbml/Unit.java
```diff
@@ -1041,182 +1041,101 @@ public static boolean isValidUnit(Model model, String unit) {
...
 
       // The resulting units is always invalid
       unit1.setKind(Kind.INVALID);
-      
-      // TODO - just set all other attributes to their default for invalid units ?
-      
-      if (!k2.equals(Kind.INVALID)) {
+
+      if (!unit2.getKind().equals(Kind.INVALID)) {
         if (unit2.isSetOffset) {
```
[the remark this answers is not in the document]
eikept: This part was just kept from the original code as LibSBML doesn't handle invalid units here. I think the idea here might have been to merge the units and only keeping the `offset`, `multiplier` and `scale`  of the valid unit. But i agree that this doesn't really make that much sense and I probably should have changed that. The question is now if these units should still be merged while also merging the parameters correctly or if they just should not be merged.

core/src/org/sbml/jsbml/Unit.java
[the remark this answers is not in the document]
eikept: So I have now looked at the ValidationTests and found that the tests in `checkConsistencyOfflineTests` all pass. My question would be if I should also test it with `OfflineValidatorTests`?  If I understand correctly I would need to set up a test folder structure with a folder for each error code for that. Is that correct?

core/src/org/sbml/jsbml/util/Maths.java
matthiaskoenig: Okay, I get it. But still don't like it. The problem you could get now is that you have possible different error accumulation depending on the exponents. I.e. if you do multiple merges with 1/3 you will have different error accumulation then with 1/2.

I still think this is overkill and in my opinion not needed. This is math on doubles and floating point errors are to be expected. The point that I by just reading the code expected the reasons for the different cases is a performance issue already shows that this is unclear and results in confusion. So if this is required please add documentation to the code for the reasons why these switch is there.

draeger: @matthiaskoenig Thank you! I am really glad you are so careful and conservative here. I think, the main use case for this specific function is merging units. There you want to have cancellation and sometimes need to come close to integer values if that is possible. Java does already provide specific functions for this, all that is done here is a case distinction followed by redirection to existing built-in functions of Java.

eikept: So I have now performed runtime analysis between the different versions of the `root()` method with `Original (König)` being the original version as it was before and `New 1 (Pertuch)` and `New 2 (Dräger)` the new versions without and with the requested changes by @draeger. Here a test data set was created containing 100 million values where every tenth value was a square root and every other tenth value was a cubic root. The rest of the values were random. Then the root function was called for each of these values either always with the same root exponent for each value (1, 2, 3, 4, 8 and 20) or with a list of root exponents for each test data value which was either completely random or one where 2s and 3s were fixed at the same position as the square/cubic roots in the test data.  The results can be seen in first of the attached figures. The original version always takes 5.7-5.9s for the complete data set. The new versions are much faster for root exponents of 1 and 2 and slightly faster for a root exponent of 3. For root exponents >4 they perform roughly the same. For the list of random exponents all algorithms perform approximately the same while the new versions are faster if some values were fixed to 2 or 3. 

Besides runtime also the average difference between the results of the original and the new algorithms was examined with the same setup as described above. The results can be seen in the second figure where only the differences between the original and the new algorithm is displayed as both new `root()` methods achieve the same results. As one would expect there was only a differences if root exponents of either 2 or 3 were used but the differences were rather small with ~10^-14 and are therefore numerically negligible.

In conclusion I would suggest that we use the `root()` method from my pull request with the changes requested by @draeger as the new method has a better runtime and can produce better results if the root exponent is either 2 or 3. I could then, as @matthiaskoenig requested, describe in the JavaDoc why the method was changed. 

runtimes_root
differences_root

matthiaskoenig: Great thanks. All my issues have been adressed. Thanks for the runtimes, this is very interesting.

core/src/org/sbml/jsbml/Unit.java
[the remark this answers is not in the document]
eikept: Since it has now been a month I wanted to ask again if somebody could answer my questions from above? Thank you in advance.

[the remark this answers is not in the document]
matthiaskoenig: Unfortunately, I don't know about the testing strategy. @niko-rodrigue and @draeger could probably answer this.

PR #247, opened 2022-05-04
Fixed a bug that can cause an OutOfBoundsException if debugging is enabled.

eikept:
**Description of Issue**

If debugging is activated in log4j and a model is read that contains a lambda element an OutOfBoundsException is thrown. The reason behind this is that during debugging a tree is drawn for each node in the model when addChild() is called on this node that also displays the unit of the node. To display the unit, the unit needs to be derived first and for lambda nodes, all children are required for this derivation. These children however are not yet present as they're being added by the addChild() method itself.

**Fix**
The code fragment that draws this tree, if debugging is enabled, was commented out.

### DIFF
--- a/core/src/org/sbml/jsbml/ASTNode.java
+++ b/core/src/org/sbml/jsbml/ASTNode.java
@@ -1821,10 +1821,19 @@ public class ASTNode extends AbstractTreeNode {
    *            the node to add as child.
    */
   public void addChild(ASTNode child) {
+    /*
+    This code for debugging was commented out because it can cause an OutOfBoundsException during
+    the parsing of a model if it contains a lambda element. The reason for this is that
+    in astNodeToTree units are derived for the node and for lambda nodes the children
+    are necessary for this derivation. The children are not yet present however as addChild()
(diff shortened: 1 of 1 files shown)


core/src/org/sbml/jsbml/ASTNode.java
```diff
@@ -1821,10 +1821,19 @@ public ASTNode(Type type, MathContainer parent) {
...
+    in astNodeToTree units are derived for the node and for lambda nodes the children
+    are necessary for this derivation. The children are not yet present however as addChild()
+    is responsible for adding them.
+     */
+    /*
     if (isDebugEnabled) {
       logger.debug(" adding child current node: \n" + astNodeToTree(this, "", ""));
       logger.debug(" adding child: \n" + astNodeToTree(child, "", ""));
     }
+     */
```
draeger, line 1836: This fixes a concrete problem but the underlying source still remains. A new [tracker item](https://github.com/sbmlteam/jsbml/issues/248) has been created to keep this active. For now, it is solved.

draeger (approved):

Issue #258, opened 2023-11-06
NullPointerException in ASTNode


draeger:
The following issue has been raised as a consequence of failing tests of the SBMLTestSuite using SBSCL (the Systems Biology Simulation Core Library). As it turns out, the problem goes back to a NPE in ASTNode, see https://github.com/draeger-lab/SBSCL/issues/74, and needs to be fixed with a new version of JSBML.

draeger:
Ah, I see: This needs to be fixed in JSBML, and a new release of that library becomes urgent in Maven for this to take effect.

dyrpsf:
> Ah, I see: This needs to be fixed in JSBML, and a new release of that library becomes urgent in Maven for this to take effect.

@draeger,
Yes, exactly – the fix is fully in JSBML (`AbstractTreeNode`/`ASTNode`).  
Once a new JSBML release is published, SBSCL will only need to update its JSBML dependency.

PR #271, opened 2026-01-06
Fix NullPointerException in AbstractTreeNode used by ASTNode clone

dyrpsf:
This PR fixes a `NullPointerException` that could occur when cloning `AbstractTreeNode` (and thus `ASTNode`) when `userObjects` is null.

Previously, the copy constructor in `AbstractTreeNode` did:

if (anode.isSetUserObjects()) {
  userObjects = new HashMap<Object, Object>();
  userObjects.putAll(anode.userObjects);
}

If a subclass overrides isSetUserObjects() (e.g. ASTNode) and returns true even when the underlying userObjects map is still null, putAll throws an NPE.

The constructor is now changed to check the field directly:

if (anode.userObjects != null && !anode.userObjects.isEmpty()) {
  userObjects = new HashMap<Object, Object>(anode.userObjects);
}

This makes the clone logic robust to such overrides.
I verified the fix by:

Building JSBML and running its tests locally.
Building SBSCL 2.1.1 against this JSBML snapshot.
Running the SBMLTestSuite wrapper used in draeger-lab/SBSCL#74 without encountering the previous NPE in ASTNode.clone.
Related issues:

Fixes sbmlteam/jsbml#258
Related to draeger-lab/SBSCL#74

### DIFF
--- a/core/src/org/sbml/jsbml/AbstractTreeNode.java
+++ b/core/src/org/sbml/jsbml/AbstractTreeNode.java
@@ -167,9 +167,10 @@ public abstract class AbstractTreeNode implements TreeNodeWithChangeSupport {
       AbstractTreeNode anode = (AbstractTreeNode) node;
       // Do not clone listeners!
       //   this.listOfListeners.addAll(anode.listOfListeners);
-      if (anode.isSetUserObjects()) {
-        userObjects = new HashMap<Object, Object>();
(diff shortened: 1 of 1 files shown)


luciansmith:
This is jsbml?  Unless I'm missing something?

dyrpsf:
Thank you for merging this!
If any further changes or adjustments are needed, please let me know.

PR #272, opened 2026-01-07
Avoid clashes in ModelBuilder by reusing existing model elements

### LINKED ISSUE #253: Add checks if a model element already exists before adding it to avoid clashes in ModelBuilder
ModelBuilder is a very convenient utility class that helps building or extending models. At the moment it assumes that elements to be added
do not exists. There could be clashes if that is not the case. Feature request: Update all creation methods in [ModelBuilder](core/src/org/sbml/jsbml/util/ModelBuilder.java) to consider that an object is already present before adding it again.

dyrpsf:
### Summary

This PR updates `ModelBuilder` so that it checks whether a model element with a given `id`
already exists before creating a new one. This avoids clashes and duplicate elements when
using `ModelBuilder` to extend or modify existing models.

The following methods now reuse existing elements if present:

- `buildModel(String id, String name)` → reuses the existing model in the `SBMLDocument` if set.
- `buildCompartment(...)` → uses `model.getCompartment(id)` if available.
- `buildParameter(...)` → uses `model.getParameter(id)` if available.
- `buildReaction(...)` → uses `model.getReaction(id)` if available.
- `buildSpecies(...)` → uses `model.getSpecies(id)` if available.
- `buildUnitDefinition(...)` → uses `model.getUnitDefinition(id)` if available.

In each case, a new element is only created if no element with the given `id` already exists.

### Rationale

Previously, `ModelBuilder` always called `createXXX(id)` on the model, assuming that elements
with the given IDs did not exist yet. When extending an existing model, this could lead to
clashes or exceptions if an element with that `id` was already present.

With these changes, `ModelBuilder` becomes safer to use for both new and existing models.

### Related issue

- Addresses [sbmlteam/jsbml#253](https://github.com/sbmlteam/jsbml/issues/253)

### DIFF
--- a/core/test/org/sbml/jsbml/util/ModelBuilderReuseTest.java
+++ b/core/test/org/sbml/jsbml/util/ModelBuilderReuseTest.java
@@ -1,0 +1,91 @@
+/*
+ * ----------------------------------------------------------------------------
+ * This file is part of JSBML. Please visit <http://sbml.org/Software/JSBML>
+ * for the latest version of JSBML and more information about SBML.
+ *
+ * Copyright (C) 2009-2022 jointly by the following organizations:
+ * 1. The University of Tuebingen, Germany
+ * 2. EMBL European Bioinformatics Institute (EBML-EBI), Hinxton, UK
+ * 3. The California Institute of Technology, Pasadena, CA, USA
+ * 4. The University of California, San Diego, La Jolla, CA, USA
+ * 5. The Babraham Institute, Cambridge, UK
+ * 
+ * This library is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation. A copy of the license agreement is provided
+ * in the file named "LICENSE.txt" included with this software distribution
+ * and also available online as <http://sbml.org/Software/JSBML/License>.
+ * ----------------------------------------------------------------------------
+ */
+
+package org.sbml.jsbml.util;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertSame;
+
+import org.junit.Test;
+import org.sbml.jsbml.Compartment;
+import org.sbml.jsbml.Model;
+import org.sbml.jsbml.SBMLDocument;
+import org.sbml.jsbml.Species;
+
+/**
+ * Tests for the reuse and clash-detection behaviour in {@link ModelBuilder}.
+ */
+public class ModelBuilderReuseTest {
+
+  /**
+   * If a compartment with the given id already exists, buildCompartment should
+   * return it unchanged and not overwrite its attributes.
+   */
+  @Test
+  public void reusesExistingCompartmentWithoutOverwritingAttributes() {
+    SBMLDocument doc = new SBMLDocument(3, 1);
+    Model model = doc.createModel("m");
+
+    Compartment original = model.createCompartment("c1");
+    original.setName("original");
+    original.setSpatialDimensions(3d);
+    original.setSize(1.23);
+    original.setUnits("litre");
+    original.setConstant(false);
+
+    ModelBuilder builder = new ModelBuilder(doc);
+
+    // Call with different arguments – these should NOT be applied,
+    // because the compartment already exists.
+    Compartment result = builder.buildCompartment("c1", true, "newName",
+        1d, 9.99, "otherUnits");
+
+    assertSame(original, result);
+    assertEquals("original", result.getName());
+    assertEquals(3d, result.getSpatialDimensions(), 0.0);
+    assertEquals(1.23, result.getSize(), 0.0);
+    assertEquals("litre", result.getUnits());
+    // remains false, not overwritten by the 'true' passed in
+    assertEquals(false, result.getConstant());
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


draeger (changes requested):
It would be great to add to the JavaDoc at least one line for each `buildXYZ` method stating that it either creates or returns the corresponding model element with the given identifier (ID). If such an element already exists, it returns that element as is without applying the other given arguments. If no such element with that ID can be found, it will be created, and the given arguments will be set as its attributes.

Please note that in the current imnplementation it is still possible that clashes can happen: In the case that another element with the given ID exists, e.g., a species with that ID when `buildCompartment` is called, the method won't find anything, but it also can't create it. Maybe it is safer to call `findNamedSBase` and check whether the returned object is an instance of the requested type.

Please note that `UnitDefinition` objects have their separate namespace.

dyrpsf:
> It would be great to add to the JavaDoc at least one line for each `buildXYZ` method stating that it either creates or returns the corresponding model element with the given identifier (ID). If such an element already exists, it returns that element as is without applying the other given arguments. If no such element with that ID can be found, it will be created, and the given arguments will be set as its attributes.
> 
> Please note that in the current imnplementation it is still possible that clashes can happen: In the case that another element with the given ID exists, e.g., a species with that ID when `buildCompartment` is called, the method won't find anything, but it also can't create it. Maybe it is safer to call `findNamedSBase` and check whether the returned object is an instance of the requested type.
> 
> Please note that `UnitDefinition` objects have their separate namespace.

Thanks for the helpful suggestions!

I’ve updated the build methods as follows:

* Each of `buildModel`, `buildCompartment`, `buildParameter`, `buildReaction`,
  `buildSpecies`, and `buildUnitDefinition` now has Javadoc explaining that the
  method either returns an existing element with the given id (without applying
  the other arguments), or creates a new one and initialises it with the
  provided attributes.

* For SId-based elements (`Compartment`, `Parameter`, `Reaction`, `Species`),
  the implementation now uses `model.findNamedSBase(id)` to detect clashes. If
  an element with the same id exists but is of a different type, an
  `IllegalArgumentException` is thrown instead of silently creating a new
  element.

* `UnitDefinition` reuses existing definitions via `model.getUnitDefinition(id)`
  and is documented as having a separate namespace, so ids can overlap with
  SId-based elements without causing clashes.

dyrpsf:
I’ve also added a small regression test class
`ModelBuilderReuseTest` that covers:

* Reusing an existing compartment without overwriting its attributes when
  `buildCompartment` is called again with the same id but different arguments.
* Throwing an `IllegalArgumentException` when attempting to build a compartment
  with an id that is already used by a `Species` (i.e. clash between types).

core/test/org/sbml/jsbml/util/ModelBuilderReuseTest.java
```diff
@@ -0,0 +1,91 @@
...
+    Species species = model.createSpecies("shared");
+    species.setCompartment("c");
+    species.setInitialConcentration(1.0);
+
+    ModelBuilder builder = new ModelBuilder(doc);
+
+    // Attempt to create a compartment with the same id.
+    // Our new implementation should detect the clash via findNamedSBase
+    // and throw an IllegalArgumentException.
+    builder.buildCompartment("shared", true, "compartment", 3d, 1.0, "litre");
```
draeger: It would be better to use the `enum` with all base units rather than writing the `String` literal `"litre"` here.

draeger (changes requested):

core/test/org/sbml/jsbml/util/ModelBuilderReuseTest.java
dyrpsf: Done! Thanks for pointing that out. Updated to use the `enum`.

dyrpsf:
Just a quick update! I have applied the requested changes to use the base unit enum instead of the string literal in the tests. Everything is pushed and the local tests are passing. Ready for another look! 🚀

core/test/org/sbml/jsbml/util/ModelBuilderReuseTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+import org.junit.Test;
+import org.sbml.jsbml.Compartment;
+import org.sbml.jsbml.Model;
+import org.sbml.jsbml.SBMLDocument;
+import org.sbml.jsbml.Species;
+import org.sbml.jsbml.Unit;
+
+/**
+ * Tests for the reuse and clash-detection behaviour in {@link ModelBuilder}.
```
draeger, line 34: You may add your `@author` tag here, @dyrpsf.

draeger (approved):

core/test/org/sbml/jsbml/util/ModelBuilderReuseTest.java
dyrpsf: Done

dyrpsf:
Thanks so much for the review and the merge, @draeger! 

I really appreciate the reminder to add the tag to the test file. I've created a quick follow-up PR to ensure my `@author` tag is attached to the class here: https://github.com/sbmlteam/jsbml/pull/304

Thanks again for all the help!

Issue #255, opened 2022-11-24
No property change happens when a value is explicitly set to NaN that was undefined before


draeger:
When listening to changes in a model, the `TreeNodeChangeListener` is activated when a previously undefined double value is explicitly set to NaN but then nothing happens because we are comparing the old value with the new value using a call to the `equals` method [here](https://github.com/sbmlteam/jsbml/blob/02c9b7db7b493105fdf690e7abe0adc76ffa3225/core/src/org/sbml/jsbml/AbstractTreeNode.java#L440). Appearantly, undefined values are NaN behind the scenes. The `setValue` method still notices the difference and calls the change listeners because there the difference is identified using an `!=` operator (see [here](https://github.com/sbmlteam/jsbml/blob/02c9b7db7b493105fdf690e7abe0adc76ffa3225/core/src/org/sbml/jsbml/QuantityWithUnit.java#L185)).

My suggestion for a fix would be to change the check in `AbstractTreeNode` so that in addition to `!equals` also a check via `!=` is performed, i.e., 
```
else if ((oldValue != null) && (!oldValue.equals(newValue) || oldValue != newValue))
```
Could such a change have any unwanted effects, too?

PR #273, opened 2026-01-09
Fix property change when undefined value is explicitly set to NaN

dyrpsf:
### Summary

This PR addresses [sbmlteam/jsbml#255](https://github.com/sbmlteam/jsbml/issues/255).

When a previously undefined double value (internally represented as `NaN`) is explicitly
set to `NaN`, no `TreeNodeChangeListener` event is fired. This happens because
`firePropertyChange` compares the old and new values using `equals()`, and for
`Double.NaN`, `oldValue.equals(newValue)` returns `true`, so no change is detected.

To fix this, the check in `AbstractTreeNode.firePropertyChange` is extended to also
consider reference inequality:

```java
if ((oldValue == null) && (newValue != null)) {
  changeType = 0; // element added
} else if ((oldValue != null) && (newValue == null)) {
  changeType = 1; // element removed
} else if ((oldValue != null) &&
           (!oldValue.equals(newValue) || oldValue != newValue)) {
  changeType = 2; // real property change
}
```
This ensures that a property change event is fired even when equals() considers
the two values equal (e.g. NaN vs NaN), but the underlying objects differ.

Verification
Built JSBML and ran the core tests using:
```BASH
mvn test -pl core -am
```

All jsbml-core tests pass, so the change does not introduce regressions
in the existing test suite.

### DIFF
--- a/core/test/org/sbml/jsbml/test/PropertyChangeNaNTest.java
+++ b/core/test/org/sbml/jsbml/test/PropertyChangeNaNTest.java
@@ -1,0 +1,67 @@
+/*
+ * ----------------------------------------------------------------------------
+ * This file is part of JSBML. Please visit <http://sbml.org/Software/JSBML>
+ * for the latest version of JSBML and more information about SBML.
+ *
+ * Copyright (C) 2009-2022 jointly by the following organizations:
+ * 1. The University of Tuebingen, Germany
+ * 2. EMBL European Bioinformatics Institute (EBML-EBI), Hinxton, UK
+ * 3. The California Institute of Technology, Pasadena, CA, USA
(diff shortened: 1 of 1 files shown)


draeger (changes requested):
Are there any test cases for the behavior of property-change events? If not, it would be good to create some. It is also possible to just create an issue for this task, so we can close this PR sooner.

dyrpsf:
I’ve added a regression test (`PropertyChangeNaNTest`) that attaches a `SimpleTreeNodeChangeListener` to a `Parameter`, then calls `setValue(Double.NaN)` when the internal value is already NaN, and verifies that a property-change event is fired.

Issue #270, opened 2025-10-28
Creators are not parsed from annotation


matthiaskoenig:
The creators are not parsed correctly for many models.
I think this is the case if the creators are written via `dcterms:creator` and with `vCard4` which the current libsbml is writing.

The old format of very old biomodels works.
```xml
<dc:creator>
<rdf:Bag>
<rdf:li rdf:parseType="Resource">
<vCard:N rdf:parseType="Resource">
<vCard:Family>Le Novère</vCard:Family>
<vCard:Given>Nicolas</vCard:Given>
</vCard:N>
<vCard:EMAIL>lenov@ebi.ac.uk</vCard:EMAIL>
<vCard:ORG rdf:parseType="Resource">
<vCard:Orgname>EMBL-EBI</vCard:Orgname>
</vCard:ORG>
</rdf:li>
</rdf:Bag>
</dc:creator>
```
But the format written by current libsbml is not parsed correctly:
```xml

<dcterms:creator>
  <rdf:Bag>
    <rdf:li rdf:parseType="Resource">
      <vCard4:hasName rdf:parseType="Resource">
        <vCard4:family-name>Elias</vCard4:family-name>
        <vCard4:given-name>Michelle</vCard4:given-name>
      </vCard4:hasName>
      <vCard4:hasEmail>michelleelias66@gmail.com</vCard4:hasEmail>
      <vCard4:organization-name>Humboldt-University Berlin, Institute for Biology</vCard4:organization-name>
    </rdf:li>
    <rdf:li rdf:parseType="Resource">
      <vCard4:hasName rdf:parseType="Resource">
        <vCard4:family-name>König</vCard4:family-name>
        <vCard4:given-name>Matthias</vCard4:given-name>
      </vCard4:hasName>
      <vCard4:hasEmail>koenigmx@hu-berlin.de</vCard4:hasEmail>
      <vCard4:organization-name>Humboldt-University Berlin, Institute for Theoretical Biology</vCard4:organization-name>
    </rdf:li>
  </rdf:Bag>
</dcterms:creator>
```

Attached an example SBML which shows the issue:

[glimepiride_liver.zip](https://github.com/user-attachments/files/23193696/glimepiride_liver.zip)

The following does not give any creators:
```java

Model model = doc.getModel();

// Check creators
History h = model.getHistory();
System.out.println(h);
for (Creator c : h.getListOfCreators()) {
    System.out.println("Creator:" + c.toString());
}
```

[...]

PR #274, opened 2026-01-12
Parse creators from dcterms:creator vCard4 annotations

dyrpsf:
### Summary

This PR extends `SBMLRDFAnnotationParser` so that creators written using the newer
`dcterms:creator` + vCard4 format (as produced by libSBML) are parsed into `History` /
`Creator` objects, in addition to the existing support for the legacy
`dc:creator` + vCard/v1 format.

Concrete changes:

- `readRDFHistory` now also looks for `dcterms:creator` elements in the `dcterms`
  namespace (`JSBML.URI_PURL_TERMS`) and calls `readCreator` on them, so both
  `<dc:creator>` and `<dcterms:creator>` blocks are handled.

- `readCreator(...)` still supports the original vCard pattern:

  - `vCard:N` → `vCard:Family`, `vCard:Given`
  - `vCard:EMAIL`
  - `vCard:ORG` → `vCard:Orgname`

  but now also falls back to vCard4 when those nodes are not present:

  - `vCard4:hasName` → `vCard4:family-name`, `vCard4:given-name`
  - `vCard4:hasEmail`
  - `vCard4:organization-name`

  The vCard4 namespace used is `http://www.w3.org/2006/vcard/ns#`.

This addresses creators that are written using vCard4 as in
[sbmlteam/jsbml#270](https://github.com/sbmlteam/jsbml/issues/270), so that
`History.getListOfCreators()` returns the expected `Creator` instances.

### Verification

- Built JSBML and ran the core tests with Java 11:

```bash
  mvn test -pl core -am
```

The jsbml-core tests pass, so the new parsing logic does not break the
existing test suite.

Fixes sbmlteam/jsbml#270

### DIFF
--- a/core/test/org/sbml/jsbml/xml/prasers/DCTermsCreatorVCard4Test.java
+++ b/core/test/org/sbml/jsbml/xml/prasers/DCTermsCreatorVCard4Test.java
@@ -1,0 +1,74 @@
+/*
+ * ----------------------------------------------------------------------------
+ * This file is part of JSBML. Please visit <http://sbml.org/Software/JSBML>
+ * for the latest version of JSBML and more information about SBML.
+ *
+ * Copyright (C) 2009-2022 jointly by the following organizations:
+ * 1. The University of Tuebingen, Germany
+ * 2. EMBL European Bioinformatics Institute (EBML-EBI), Hinxton, UK
+ * 3. The California Institute of Technology, Pasadena, CA, USA
+ * 4. The University of California, San Diego, La Jolla, CA, USA
+ * 5. The Babraham Institute, Cambridge, UK
+ * 
+ * This library is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


dyrpsf:
> I don't see if a test case was created. It would be better to add one.

Thanks for the feedback! I’ve added a regression test for the new dcterms:creator + vCard4 handling.

New test + resource:
* `core/test/org/sbml/jsbml/xml/parsers/DCTermsCreatorVCard4Test.java`
* `core/test/org/sbml/jsbml/xml/parsers/dcterms-creator-vcard4-example.xml`

The test loads a model annotated with `dcterms:creator` in the vCard4 format and checks that
`History.getListOfCreators()` returns two Creator objects with the expected given name, family
name, and organisation (Given1/Family1/Org1 and Given2/Family2/Org2).

dyrpsf:
> Well done!

Issue #252, opened 2022-10-13
Cloning of GeneProduct may create invalid SBML because of missing required attribute `label`


draeger:
The clone [constructor](https://github.com/sbmlteam/jsbml/blob/02c9b7db7b493105fdf690e7abe0adc76ffa3225/extensions/fbc/src/org/sbml/jsbml/ext/fbc/GeneProduct.java#L119) in `GeneProduct` only calls its superconstructor. In addition, it should also copy all of its attributes, `label` in particular (which is a required attribute). As of now, cloning gene products results in invalid SBML.

dyrpsf:
Hi @draeger ,
This issue is now addressed in PR #275.

The root cause was that GeneProduct.clone() delegates to the copy constructor, but the GeneProduct(GeneProduct nsb) constructor only called super(nsb) and did not copy the required label and associatedSpecies fields. As a result, cloned GeneProduct instances could be invalid and lead to malformed SBML.

The fix updates the copy constructor to explicitly copy both fields when they are set:
```java
if (nsb.isSetLabel()) {
    setLabel(nsb.getLabel());
}
if (nsb.isSetAssociatedSpecies()) {
    setAssociatedSpecies(nsb.getAssociatedSpecies());
}

```
With this change, clone() now produces a fully equivalent GeneProduct, preserving all required metadata and preventing invalid SBML generation.

Once PR #275 is merged, this issue can be safely closed.

PR #275, opened 2026-01-13
Fix GeneProduct cloning to copy required label and associatedSpecies

dyrpsf:
### Summary

This PR fixes [sbmlteam/jsbml#252](https://github.com/sbmlteam/jsbml/issues/252).

Previously, the `GeneProduct(GeneProduct nsb)` copy constructor only called the
super-constructor and did not copy the `label` (required) or `associatedSpecies`
fields. Since `clone()` is implemented as:

```java
@Override
public GeneProduct clone() {
  return new GeneProduct(this);
}
```
cloning a GeneProduct would result in a cloned instance missing these fields,
potentially producing invalid SBML.

The copy constructor is now:
```java
public GeneProduct(GeneProduct nsb) {
  super(nsb);

  if (nsb.isSetLabel()) {
    setLabel(nsb.getLabel());
  }

  if (nsb.isSetAssociatedSpecies()) {
    setAssociatedSpecies(nsb.getAssociatedSpecies());
  }
}
```

so that both label and associatedSpecies are preserved when cloning.

Verification
Local compilation with Maven (core + fbc) succeeds using Java 11.
Cloning a GeneProduct now yields an equal instance with the same label
and associatedSpecies.

Fixes sbmlteam/jsbml#252

### DIFF
--- a/extensions/fbc/src/org/sbml/jsbml/ext/fbc/GeneProduct.java
+++ b/extensions/fbc/src/org/sbml/jsbml/ext/fbc/GeneProduct.java
@@ -112,12 +112,22 @@ public class GeneProduct extends AbstractNamedSBase implements UniqueNamedSBase 
   }
 
   /**
-   * Creates a new {@link GeneProduct} instance.
+   * Creates a new {@link GeneProduct} instance by copying all fields.
    * 
    * @param nsb the instance to clone
    */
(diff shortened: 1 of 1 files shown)


dyrpsf:
Thanks for merging this PR! I am deleting the branch...
If any further changes or adjustments are needed, please let me know.

PR #276, opened 2026-01-14
Fix OverdeterminationValidator for L3V2 rateOf csymbol

### LINKED ISSUE #138: correct OverdeterminationValidator for L3V2
In L3V2, you can have inside the mathML the csymbol rateOf on top of reaction id. The current code does not take this into account and the validation is not properly detecting problems from some L3V2 models.

dyrpsf:
###Summary

Fixes #138 .

In SBML Level 3 Version 2, rateOf csymbols can target a reaction identifier. The OverdeterminationValidator was not correctly taking these references into account because, in some cases, the ASTNode for the identifier has no variable set, and the identifier was therefore ignored when building the bipartite graph. This could cause overdetermined models involving rateOf(reactionId) to go undetected.

Changes

1. OverdeterminationValidator

- Added a helper:

```Java

private SBase resolveSBaseFromId(String id)
```
which resolves an identifier string to the corresponding SBase in the model (Species, Compartment, Parameter, or Reaction).

- Updated

```Java

private void getVariables(ListOf<LocalParameter> param,
                          ASTNode node,
                          List<SBase> variables,
                          int level)
```
to:

First use node.getVariable() as before.
If getVariable() returns null (which can happen for identifiers used under a rateOf csymbol in L3V2), fall back to resolveSBaseFromId(node.getName()).
Preserve existing behavior for:
excluding local parameters

- special Level 1 parameter insertion ordering

- ignoring time and Avogadro symbols.

- As a result, algebraic rules whose MathML uses rateOf(reactionId) now correctly link to the corresponding reaction node in the validator’s bipartite graph.

2. Tests

- Added core/test/org/sbml/jsbml/validator/OverdeterminationValidatorTest.java.

- New test:

```Java

testGetVariablesResolvesReactionIdWhenVariableNotSet()
```

- Builds a minimal L3V2 model with a single reaction R1.

- Creates an ASTNode that refers to "R1" by name, without setting its variable (mimicking the situation for rateOf targets).

- Invokes the private getVariables(...) method via reflection.

- Asserts that the resulting variable list contains the reaction R1.

Testing

On this branch I ran:

- mvn -pl core -am -DskipTests install

- mvn -pl core test

Both complete with BUILD SUCCESS. No changes were made to project POMs or external dependencies.

### DIFF
--- a/core/src/org/sbml/jsbml/validator/OverdeterminationValidator.java
+++ b/core/src/org/sbml/jsbml/validator/OverdeterminationValidator.java
@@ -38,6 +38,7 @@ import org.sbml.jsbml.Compartment;
 import org.sbml.jsbml.ListOf;
 import org.sbml.jsbml.LocalParameter;
 import org.sbml.jsbml.Model;
+import org.sbml.jsbml.NamedSBase;
 import org.sbml.jsbml.Parameter;
 import org.sbml.jsbml.RateRule;
 import org.sbml.jsbml.Reaction;
@@ -45,6 +46,9 @@ import org.sbml.jsbml.Rule;
 import org.sbml.jsbml.SBase;
 import org.sbml.jsbml.Species;
 import org.sbml.jsbml.SpeciesReference;
+import org.w3c.dom.Node;
+
+import com.thoughtworks.xstream.io.binary.Token.StartNode;
 
 /**
  * This class creates a bipartite graph and a matching for the given model using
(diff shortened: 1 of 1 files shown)


dyrpsf:
> Great work! Thanks for this contribution!

Thank you! I appreciate the review and the feedback. Glad I could contribute. Looking forward to working on more improvements in the future.

Issue #251, opened 2022-08-02
Improve flattening by testing and merging an existing fork


draeger:
The flattening algorithm implemented in JSBML should be improved. A better version has been implemented in the form of a [fork](https://github.com/eikept/jsbml). It is nearly ready but requires more extensive testing before a pull request to the master branch can be issued.

luciansmith:
There are test models in the libsbml source you would be welcome to use; a few at

https://github.com/sbmlteam/libsbml/tree/development/src/sbml/packages/comp/extension/test/test-data

and many more at

https://github.com/sbmlteam/libsbml/tree/development/src/sbml/packages/comp/util/test/test-data

dyrpsf:
Hi @draeger & @luciansmith ,
I’ve opened a PR with the flattening updates from the `eikept/jsbml` fork:
https://github.com/sbmlteam/jsbml/pull/277

Summary:

- Ported `CompFlatteningConverter` and its helper classes (`CompFlatteningIDExpander`,
  `CompFlatteningIDExpanderTS`) from the fork into `extensions/comp`.
- Updated the merge order in `mergeModels` so that the parent model’s objects appear
  before the flattened submodels’ objects, which makes the flattened output match
  the existing `testFlattening/testN_flat.xml` expectations (e.g. for test6).
- Ran `CompFlattenTest` on the local comp test data; the individual flattening and
  internalisation tests now pass with the new converter.

Remaining points:

- `testAllData` still reports one mismatch somewhere in the full `testFlattening`
  suite (this was already marked in the test with a TODO).
- `testInternaliseExternalModelDefinitions_online` continues to fail because the
  URL `https://sbmlteam.github.io/jsbml/test/data/comp/simple_online_url.xml`
  returns a 301 HTML page instead of SBML, which seems like a test-data/hosting
  issue rather than a flattening bug.

Feedback on the approach and any preferences for handling the remaining tests would
be very welcome.

PR #277, opened 2026-01-16
Port improved comp flattening from eikept/jsbml and run tests

dyrpsf:
This PR addresses part of #251 by porting the improved comp flattening implementation
from the `eikept/jsbml` fork and wiring it into the current `jsbml-comp` module.

**Main changes**

- Replaced `CompFlatteningConverter` with the version from the `eikept/jsbml` fork,
  which provides a more complete flattening implementation for the comp package
  (including proper handling of ReplacedElement/Deletion chains, conversion factors,
  and nested submodels).
- Added the helper classes used by the new converter:
  - `CompFlatteningIDExpander`
  - `CompFlatteningIDExpanderTS`
- Adjusted the `mergeModels` method to merge the “current” (parent) model before
  the previously-flattened submodels. This preserves the original parameter ordering,
  so that flattened output matches the existing `testFlattening/testN_flat.xml`
  expectations (e.g. for test6).

**Testing**

From this branch I ran:

```bash
mvn -pl :jsbml-comp -Dtest=CompFlattenTest#testSpecificFile test
mvn -pl :jsbml-comp -Dtest=CompFlattenTest test
```
Results:

- testSpecificFile and the individual flattening/internalisation tests now pass for the local testFlattening and testGathering data.

- testAllData now exercises the full testFlattening suite; there is still one mismatch reported via Success Testing Model (this was previously flagged in the test with a TODO comment).

- testInternaliseExternalModelDefinitions_online still errors due to the remote URL (https://sbmlteam.github.io/jsbml/test/data/comp/simple_online_url.xml) returning a 301 HTML redirect instead of SBML; this looks like a test-data/ hosting issue rather than an algorithm problem.

I’ve kept those two behaviours unchanged but can help adjust the tests if you
prefer (e.g. updating the URL to a raw SBML resource or marking the online test
as integration-only).

### DIFF
--- a/extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
+++ b/extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
@@ -41,22 +41,22 @@ public class CompFlatteningConverter {
 
     private ListOf<ModelDefinition> modelDefinitionListOf;
 
-    // HashMap containing prefix for each submodel
-    private HashMap<List<String>, String> subModelPrefixes;
+    // Map containing prefix for each submodel
+    private Map<List<String>, String> subModelPrefixes;
 
     private List<Submodel> listOfSubmodelsToFlatten;
 
-    // Hashmap of Replaced Elements in the form pathToModel -> idType (id, metaID, ...) -> replacedId -> replacedElementInfo
-    private HashMap<List<String>, HashMap<IdType, HashMap<String, ReplacedElementInfo>>> replacedAndDeletedElementsHashMap;
+    // Map of Replaced Elements in the form pathToModel -> idType (id, metaID, ...) -> replacedId -> replacedElementInfo
+    private Map<List<String>, Map<IdType, Map<String, ReplacedElementInfo>>> replacedAndDeletedElementsHashMap;
 
     // Contains newly created (generally copied) initial assignments
-    private HashMap<List<String>, List<InitialAssignment>> newInitAssignHashMap;
+    private Map<List<String>, List<InitialAssignment>> newInitAssignHashMap;
 
     // Contains newly created (generally copied) rules
-    private HashMap<List<String>, List<Rule>> newRulesHashMap;
+    private Map<List<String>, List<Rule>> newRulesHashMap;
 
     // Contains replaced units (their references can only be changed at the end)
-    private HashMap<String, String> newUnitDefMaps;
+    private Map<String, String> newUnitDefMaps;
 
     // Final flattened model that is merged into bottom-up in the submodel tree
     private Model flattenedModel;
@@ -65,7 +65,7 @@ public class CompFlatteningConverter {
     private List<String> curPath;
 
     // Previously visited models
-    private HashMap<String, Model> visitedModels;
+    private Map<String, Model> visitedModels;
 
     public CompFlatteningConverter() {
         this.listOfSubmodelsToFlatten = new ArrayList<>();
@@ -223,7 +223,7 @@ public class CompFlatteningConverter {
         parseAllReplacedBy(modelCopy, this.curPath);
 
         // STEP 3: Remove all objects which were either deleted or replaced
-        HashMap<String, ReplacedElementInfo> replacedElements = removeReplacedAndDeletedElements(modelCopy, compModelPlugin);
+        Map<String, ReplacedElementInfo> replacedElements = removeReplacedAndDeletedElements(modelCopy, compModelPlugin);
 
         // STEP 4: Add newly created InitialAssignments and Rules
         addNewElementsWithVariables(modelCopy);
@@ -259,7 +259,7 @@ public class CompFlatteningConverter {
      */
     private void parseListOfDeletions(List<Deletion> deletions, List<String> curPath, String submodelID) {
 
-        HashMap<String, ReplacedElementInfo> replacedIDElementsInModel = null;
+        Map<String, ReplacedElementInfo> replacedIDElementsInModel = null;
         if(this.replacedAndDeletedElementsHashMap.containsKey(curPath)) {
             replacedIDElementsInModel = this.replacedAndDeletedElementsHashMap.get(curPath).get(IdType.ID);
         }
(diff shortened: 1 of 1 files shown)


extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
```diff
@@ -19,72 +19,75 @@
...
 
     private final static Logger LOGGER = Logger.getLogger(CompFlatteningConverter.class.getName());
 
-    private List<String> previousModelIDs;
-    private List<String> previousModelMetaIDs;
     private ListOf<ModelDefinition> modelDefinitionListOf;
-    private ListOf<ExternalModelDefinition> externalModelDefinitionListOf;
+
+    // HashMap containing prefix for each submodel
+    private HashMap<List<String>, String> subModelPrefixes;
```
draeger: Do we have to declare it as a `HashMap` or would a `Map` (interface) be sufficient?

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
```diff
@@ -19,72 +19,75 @@
...
+    private HashMap<String, String> newUnitDefMaps;
+
+    // Final flattened model that is merged into bottom-up in the submodel tree
     private Model flattenedModel;
 
+    // Path to currently visited submodel (node)
+    private List<String> curPath;
+
+    // Previously visited models
+    private HashMap<String, Model> visitedModels;
```
draeger: The comment above applies to all other instances of `HashMap`, too. I'd use `Map` whenever possible.

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
```diff
@@ -19,72 +19,75 @@
...
+
+        this.newUnitDefMaps = new HashMap<>();
 
         this.flattenedModel = new Model();
 
+        this.subModelPrefixes = new HashMap<List<String>, String>();
+
+        this.curPath = new ArrayList<>();
+
+        this.visitedModels = new HashMap<>();
```
draeger, line 89: Of course, initializing them as `HashMap` is a separate story. Declare as `Map` and then decide what is the most useful implementation for the purpuse.

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
```diff
@@ -103,211 +106,1187 @@ public SBMLDocument flatten(SBMLDocument document) {
...
+                returnValue = selectCorrectRef(finalSBaseRef);
+            } else {
+                returnValue = selectCorrectRef(deletion);
+            }
+            deletedElementRef = returnValue.getKey();
+            deletedElementType = returnValue.getValue();
+
+            // Add deleted element to replacedAndDeletedElementsHashMap
+            if (deletedElementRef != null) {
+                HashMap<IdType, HashMap<String, ReplacedElementInfo>> deletedElementsForModel = this.replacedAndDeletedElementsHashMap.get(deletionPath);
```
draeger: Also here: Please use `Map` unless we must declare a `HashMap`.

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
```diff
@@ -103,211 +106,1187 @@ public SBMLDocument flatten(SBMLDocument document) {
...
+     *
+     * @param node
+     * @param parent
+     * @param replacedElementsHashMap
+     * @param subModelPrefix
+     * @param timeConvFactorNode
+     * @param i
+     */
+    private void applyConvFactorToNode(ASTNode node, ASTNode parent,
+                                       HashMap<String, ReplacedElementInfo> replacedElementsHashMap,
```
draeger: In particular when declaring a method (even a private one) it is better to declare the parameter using the most generic but as specific as needed type, in this case most likely a `Map` rather than a specific `HashMap` implementation.

dyrpsf:
Hi @draeger ,

I have refactored `CompFlatteningConverter.java` to use generic interfaces (`Map`, `List`) for field declarations and method parameters as requested.

I noticed that `OverdeterminationValidator.java` and its corresponding test also use `HashMap` and `ArrayList`. Would you like me to apply the same type-level refactoring to those files as well, or should we keep this PR focused strictly on the flattening converter?

The project compiles successfully, though some existing network-related timeouts persist in `TestExternalModelDefinition`.

Thanks!

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
dyrpsf: Done. Changed the declaration to Map.

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
dyrpsf: Done. Updated the method signature to use Map instead of HashMap.

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
dyrpsf: Acknowledged. I've updated all class fields and private methods in this file to use generic interfaces.

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
dyrpsf: Done. I have updated the field declarations at the top of the class to use the List and Map interfaces, while keeping the specific HashMap/ArrayList initializations here in the constructor.

extensions/comp/src/org/sbml/jsbml/ext/comp/util/CompFlatteningConverter.java
dyrpsf: Done. I have refactored the local variable declarations in this method to use the generic Map interface.

dyrpsf:
> Thanks for your hard work, @dyrpsf!

Thank you for merging this and for the kind words! If any further changes or adjustments are needed, please let me know.

PR #280, opened 2026-02-01
Handle empty lambda expressions in UnitsCompiler (fix #248)

### LINKED ISSUE #248: Deriving a unit for a lambda expression may cause an out of bounds exception
Lambda expressions are placeholders that are filled with arguments when the expression is called. Consequently, their units can in many cases not directly be derived if arguments are passed to the expression. The reason is that SBML does not restrict assignable types in lambda expressions. Consequently, various objects (such as species, compartments, or parameters) could be passed to a lambda expression when a function is called. If used in muliple places, the same function may result in varying derived units. In its current form, the attempt to derive units of a lambda expression itself (i.e., not within a call hierarchy but the structure itself) may result in an out of bounds exception because the lambda expression itself does not know which arguments are possibly passed to it. Those have to be filled in when the expression is called. More specifically, the in methods such as `astNodeToTree` wihtin `ASTNode` units are derived for the node and for lambda nodes the children are necessary for this derivation. The children are not present at that time, however, because `addChild()` is responsible for adding them.

Consequently, the derivation of units from lambda expressions in the corresponding units compiler needs to be revised.

dyrpsf:
Fixes #248.

Deriving units for a lambda expression while its child list is still empty
(e.g., immediately after the lambda node is constructed, before `addChild()`
has been called) could cause an `IndexOutOfBoundsException` in
`UnitsCompiler.lambda(List<ASTNode2>)`.

Previously, `lambda(...)` always did:

```java
values.get(values.size() - 1)
```
without checking whether values was empty, so an empty child list led to a
call to `values.get(-1)`.

Changes

1. Updated UnitsCompiler.lambda(List<ASTNode2> values) to guard against an empty list:

    - If `values` is `null` or empty, the method now returns `invalid()`, i.e., an `ASTNode2Value` whose `UnitDefinition` is marked as `INVALID`. This avoids out-of-bounds access while signalling that units for a bare lambda expression (without arguments/body yet) are unknown at that time.

2. For non-empty `values`, behaviour is unchanged:

    - All but the last child are treated as arguments and stored in `namesToUnits`.

    - The units of the last child (the body) are used as the lambda’s result units.

Tests

1. Added UnitsCompilerLambdaTest under `core/test/org/sbml/jsbml/math/compiler`:

    - `testLambdaWithNoChildrenDoesNotThrow()`:

        - Calls new `UnitsCompiler(3, 1).lambda(Collections.<ASTNode2>emptyList())`.

        - Asserts that no `IndexOutOfBoundsException` is thrown.

        - Asserts that the returned units are marked as invalid.

2. Ran:

```Bash
mvn -pl core -Dtest=UnitsCompilerLambdaTest test
mvn -pl core test
```
Both complete with `BUILD SUCCESS` on Java 11.

### DIFF
--- a/core/src/org/sbml/jsbml/math/compiler/UnitsCompiler.java
+++ b/core/src/org/sbml/jsbml/math/compiler/UnitsCompiler.java
@@ -727,12 +727,23 @@ public class UnitsCompiler implements ASTNode2Compiler {
   @Override
   // TODO: Specify generic type T i.e. ASTNode2Value<?>
   public <T> ASTNode2Value<?> lambda(List<ASTNode2> values) throws SBMLException {
-    for (int i = 0; i < values.size() - 1; i++) {
-      namesToUnits.put(values.get(i).toString(),
-        values.get(i).compile(this));
+    // When the lambda node is first constructed, its child list may be empty;
+    // in that case, there is no body and no arguments yet, so we cannot derive
+    // any meaningful units. Returning an invalid unit definition here avoids
+    // out-of-bounds exceptions while signalling "units unknown".
+    if (values == null || values.isEmpty()) {
+      return invalid();
     }
-    return new ASTNode2Value(values.get(values.size() - 1).compile(this)
-      .getUnits(), this);
+
+    // For a fully-formed lambda expression, all but the last child are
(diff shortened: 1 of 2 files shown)


dyrpsf:
> Well done!

Issue #246, opened 2022-04-28
Make offline validation the default


draeger:
JSBML's validation capabilities have greatly improved and are now very reliable. In some situations, the online validation can be problematic, e.g., if no network connection is available or when the URL pointing to the service needs to change. The latter problem could be fixed by implementing a configurable solution in an exchangeable file, but switching the default behavior of JSBML's validation to offline mode would ease many processes.

draeger:
Checking PR #281.

PR #281, opened 2026-02-02
Document offline validation as the default and add test (#246)

dyrpsf:
This PR addresses sbmlteam/jsbml#246.

## Summary

`SBMLDocument.checkConsistency()` already delegates to `checkConsistencyOffline()`, so offline validation is effectively the default. However, the Javadoc still claimed that validation was "currently performed using the SBML.org online validator". This PR:

* updates the documentation to match the actual behaviour, and
* adds a small regression test to ensure `checkConsistency()` continues to use the offline validator.

## Changes

1. **Javadoc updates in `SBMLDocument`**
   * `core/src/org/sbml/jsbml/SBMLDocument.java`
     * `checkConsistencyOffline()` Javadoc now explains that the offline validator is used by default when calling `checkConsistency()`, and points to `checkConsistencyOnline()` as the alternative.
     * `checkConsistency()` Javadoc is updated to state that validation is performed using the JSBML internal offline validator, with the SBML.org online validator still available via `checkConsistencyOnline()`.

   > Note: No executable code was changed; only comments/Javadoc were modified.  
   > The method remains:
   > ```java
   > public int checkConsistency() {
   >   return checkConsistencyOffline();
   > }
   > ```

2. **New unit test**

   * Added `core/test/org/sbml/jsbml/test/SBMLDocumentValidationModeTest.java`
   * The test creates a small `TestDocument` subclass of `SBMLDocument` that overrides:
     * `checkConsistencyOffline()` and
     * `checkConsistencyOnline()`
     and records which one is called.
   * The test `checkConsistencyUsesOfflineValidatorByDefault()` asserts that:
     * `checkConsistency()` calls the offline variant,
     * does **not** call the online variant,
     * and returns the offline method’s result.

   This guards against future regressions where `checkConsistency()` might accidentally call the online validator again.

## Tests

Executed locally:

* `mvn -pl core test` ✅

Test Results

The newly added regression test behaves as expected and passes successfully.

### DIFF
--- a/core/src/org/sbml/jsbml/SBMLDocument.java
+++ b/core/src/org/sbml/jsbml/SBMLDocument.java
@@ -336,9 +336,9 @@ public class SBMLDocument extends AbstractSBase {
    * flag in the individual SBMLError objects returned by
    * {@link SBMLDocument#getError(int)} to determine the nature of the failures.
    * 
-   * <p>The offline validator is not yet as complete as the online one so it is not
-   * currently used by default when calling {@link #checkConsistency()} but it might be
-   * in future versions. To get an up to date status, please check the page 
+   * <p>The offline validator was initially less complete than the online one but
+   * has evolved and is now used by default when calling {@link #checkConsistency()}.
+   * The online validator remains available via {@link #checkConsistencyOnline()}. 
    * <a href="https://github.com/sbmlteam/jsbml/wiki/Offline-validator-status">Offline-validator-status</a>.</p>
    * 
    * @return the number of errors found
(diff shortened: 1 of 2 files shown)


ntung:
It's very helpful work! Thank you @dyrpsf  The BioModels repository is struggling with the consistency check.

dyrpsf:
> It's very helpful work! Thank you @dyrpsf The BioModels repository is struggling with the consistency check.

Thank you for the feedback and for explaining the BioModels use case! I’m glad this change helps with your consistency checks. If you run into any other issues or notice cases where the offline validator behaves unexpectedly, I’d be happy to look into follow‑up improvements.

core/src/org/sbml/jsbml/SBMLDocument.java
```diff
@@ -580,9 +580,11 @@ public int checkConsistencyOnline() {
 
   /**
    * Validates the {@link SBMLDocument}.
-   * 
-   * <p> The validation is currently performed using the
-   * SBML.org online validator (http://sbml.org/validator/).</p>
+   *
+   * <p>The validation is currently performed using the JSBML internal
```
draeger: Please change “currently“ to “primarily”.

dyrpsf:
I noticed that the “currently” → “primarily” wording change mentioned in the review wasn’t yet in master, so I opened a small follow-up PR https://github.com/sbmlteam/jsbml/pull/287 to adjust that sentence in the Javadoc.

Issue #264, opened 2024-12-10
org.sbml.jsbml.xml.XMLNode cannot be cast to org.sbml.jsbml.Annotation


ntung:
Dear The JSBML Developers Team,

BioModels has some curated models with the issue. These models were curated for a while but they run into this problem. The problem broke down our index pipeline. Would you be able to escalate the problem? 
BIOMD0000000596, BIOMD0000000794, BIOMD0000000849, BIOMD0000000925, BIOMD0000000965. 
```
java.lang.ClassCastException: org.sbml.jsbml.xml.XMLNode cannot be cast to org.sbml.jsbml.Annotation
	at org.sbml.jsbml.xml.stax.SBMLReader.readXMLFromXMLEventReader(SBMLReader.java:851)
	at org.sbml.jsbml.xml.stax.SBMLReader.readSBMLFromStream(SBMLReader.java:554)
	at org.sbml.jsbml.xml.stax.SBMLReader.readSBMLFromStream(SBMLReader.java:564)
	at org.sbml.jsbml.SBMLReader.readSBMLFromStream(SBMLReader.java:251)
	at org.sbml.jsbml.SBMLReader$readSBMLFromStream.call(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:47)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:116)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:128)
	at net.biomodels.jummp.indexing.indexer.SBMLIndexer.extractFileContent(SBMLIndexer.groovy:129)
```
I have seen another user reporting the problem at #263.

Schmoho:
Hi! I am not a JSBML dev, but I looked into your issue because I am also encountering it.
I only took a look at BIOMD0000000965, but I guess the issue will be more or less the same:

The top-level annotation (on the `<sbml>` element) of this file is invalid SBML, because the layout-information is within the annotation object. This looks like a line-swap, likely either introduced by COPASY or someone manually fidgeting with the XML. Moving the `<layout:listOfLayouts>` tag out of the `<annotation>` tag "solves" the issue apparently.

However, I would still consider this a bug in JSBML; at the very least, this should have a more informative error message.

Schmoho:
I just realized, the SBML is not actually invalid:
[...]

Issue #263, opened 2024-10-29
ClassCastException in SBMLReader


danv61:
The attached xml file triggers the following exception:

java.lang.ClassCastException: class org.sbml.jsbml.xml.XMLNode cannot be cast to class org.sbml.jsbml.Annotation (org.sbml.jsbml.xml.XMLNode and org.sbml.jsbml.Annotation are in unnamed module of loader 'app')

	at org.sbml.jsbml.xml.stax.SBMLReader.readXMLFromXMLEventReader(SBMLReader.java:851)

The following warning was also generated:
WARN (SBMLReader.java:735) - Starting to read a new annotation element while the previous annotation element is not finished.

Here is the code line in question:
              annoReader.processAnnotation((SBase) ((Annotation) lastElement).getParent()); // or take the second element in the stack ??

I checked the file with the online SBML validator and it is valid.
[Ota2015_GDI-integrated.zip](https://github.com/user-attachments/files/17560504/Ota2015_GDI-integrated.zip)

Schmoho:
Hi! I just looked into the file and it seems to be the same issue as the related  #264 

Unfortunately without deeper understanding of your use case it is pretty hard to say what the correct course of action for you would be here.

PR #282, opened 2026-02-13
Guard SBMLReader annotation processing against ClassCastException (#264)

dyrpsf:
This PR addresses sbmlteam/jsbml#264.

## Problem

For some SBML models (e.g. certain BioModels/COPASI exports), reading the file with JSBML fails with:

```text
java.lang.ClassCastException: org.sbml.jsbml.xml.XMLNode cannot be cast to org.sbml.jsbml.Annotation
    at org.sbml.jsbml.xml.stax.SBMLReader.readXMLFromLVLEventReader(SBMLReader.java:851)
```

The discussion on #264 and #263 indicates that this occurs when the model annotation contains nested layout elements, e.g. a `layout:listOfLayouts` inside `<model><annotation>…</annotation></model>`.

Root cause
In the StAX-based reader:

`core/src/org/sbml/jsbml/xml/stax/SBMLReader.java`

within the `event.isEndElement()` handling, at the end of an `<annotation>` element, the code did:

```java
if (currentNode.getLocalPart().equals("annotation") && isSBMLelement)
{
  isInsideAnnotation = false;
  annotationDeepness = -1;

  // calling the annotation parsers
  for (AnnotationReader annoReader : annotationParsers) {
    annoReader.processAnnotation((SBase) ((Annotation) lastElement).getParent()); // or take the second element in the stack ??
  }
}
```
This unconditionally casts `lastElement` to `Annotation` and its parent to `SBase`.
If, for some models, the top of the internal stack is an `XMLNode` instead of an `Annotation`, this cast fails with the reported `ClassCastException`, aborting parsing.

Several comments on `#264` suggested that JSBML should instead log an error and continue, checking the type before the cast.

Fix
In `SBMLReader`, the `annotation` end-element handling is now guarded by type checks:

```java
if (currentNode.getLocalPart().equals("annotation") && isSBMLelement)
{
  isInsideAnnotation = false;
  annotationDeepness = -1;

  // calling the annotation parsers, but be robust against unexpected stack contents
  if (lastElement instanceof Annotation) {
    Annotation annotation = (Annotation) lastElement;
    Object parent = annotation.getParent();

    if (parent instanceof SBase) {
      for (AnnotationReader annoReader : annotationParsers) {
        annoReader.processAnnotation((SBase) parent); // or take the second element in the stack ??
      }
    } else {
      logger.error("End of <annotation>: expected parent of Annotation to be an SBase but found "
          + (parent == null ? "null" : parent.getClass().getCanonicalName())
          + ". Skipping annotation parsing.");
    }
  } else {
    logger.error("End of <annotation>: expected top stack element to be an Annotation but found "
        + (lastElement == null ? "null" : lastElement.getClass().getCanonicalName())
        + ". Skipping annotation parsing.");
  }
}
```

Behavior:

  - Valid case (unchanged):

    - If `lastElement` is an `Annotation` and its parent is an `SBase`, `AnnotationReader.processAnnotation(...)` is called exactly as before.

  - Invalid / unexpected case (was crashing before):

    - If `lastElement` is not an `Annotation`, or its parent is not an `SBase`, JSBML now:

      - logs an error via the existing `Logger`, and

      - skips calling the annotation parsers for this annotation,

      - but continues parsing the rest of the SBML document.

Thus we avoid the `ClassCastException` while still surfacing that something is off in the annotation structure.

New test
Added a regression-style test:

- `core/test/org/sbml/jsbml/test/SBMLReaderAnnotationCastTest.java`

```java
public class SBMLReaderAnnotationCastTest {

  @Test
  public void layoutInModelAnnotationDoesNotThrow() throws Exception {
    String xml =
        "<sbml xmlns=\"http://www.sbml.org/sbml/level2/version4\" " +
        "      xmlns:layout=\"http://www.sbml.org/sbml/level3/version1/layout/version1\" " +
        "      level=\"2\" version=\"4\">" +
        "  <model id=\"m\">" +
        "    <annotation>" +
        "      <layout:listOfLayouts>" +
        "        <layout:layout id=\"L1\"/>" +
        "      </layout:listOfLayouts>" +
        "    </annotation>" +
        "  </model>" +
        "</sbml>";

    SBMLDocument doc = new SBMLReader().readSBMLFromString(xml);

    // If we reach here, no ClassCastException (or other exception) was thrown.
    assertNotNull(doc);
    assertNotNull(doc.getModel());
  }
}
```

This exercises an SBML L2V4 document whose model annotation contains a `layout:listOfLayouts` element, mirroring the situation described in #264, and verifies that `readSBMLFromString(...)` returns a document with a model (i.e., no exception was thrown).

Tests
Executed locally:

- `mvn -pl core test` ✅

I do not have the exact `BioModels` files mentioned in #264 on this machine, but this change narrowly guards the cast that previously failed and adds a unit test to cover a representative layout-in-annotation scenario.

### DIFF
--- a/core/src/org/sbml/jsbml/xml/stax/SBMLReader.java
+++ b/core/src/org/sbml/jsbml/xml/stax/SBMLReader.java
@@ -34,6 +34,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Stack;
 
+
 import javax.xml.namespace.QName;
 import javax.xml.stream.XMLEventReader;
 import javax.xml.stream.XMLInputFactory;
@@ -75,6 +76,8 @@ import org.sbml.jsbml.xml.parsers.XMLNodeReader;
 
 import com.ctc.wstx.api.WstxInputProperties;
 import com.ctc.wstx.stax.WstxInputFactory;
+
+import static java.text.MessageFormat.format;
 
 
 /**
@@ -856,14 +859,16 @@ public class SBMLReader {
                   annoReader.processAnnotation((SBase) parent); // or take the second element in the stack ??
                 }
               } else {
-                logger.error("End of <annotation>: expected parent of Annotation to be an SBase but found "
-                    + (parent == null ? "null" : parent.getClass().getCanonicalName())
-                    + ". Skipping annotation parsing.");
+                logger.error(format(
+                    "End of <annotation>: expected parent of Annotation to be an SBase but found {0}. Skipping annotation parsing.",
+                    parent == null ? "null" : parent.getClass().getCanonicalName()
+                ));
               }
             } else {
-              logger.error("End of <annotation>: expected top stack element to be an Annotation but found "
-                  + (lastElement == null ? "null" : lastElement.getClass().getCanonicalName())
-                  + ". Skipping annotation parsing.");
+              logger.error(format(
+                  "End of <annotation>: expected top stack element to be an Annotation but found {0}. Skipping annotation parsing.",
+                  lastElement == null ? "null" : lastElement.getClass().getCanonicalName()
+              ));
             }
             
           } else if (isInsideAnnotation) {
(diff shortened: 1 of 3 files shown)


dyrpsf:

PR #290, opened 2026-03-31
Fix #129: Implement Arrays Validation Rule 20116 for SBaseRef missing indices

### LINKED ISSUE #129: Bug: Arrays Validation Rule 20116
Arrays validation needs to be updated. It is not checking validation rules correctly. A major problem is when SBaseRef points to an array and the object doesn't have index objects. The validator is not finding the expected error, which should indicate SBaseRef should deference to a scalar object.

dyrpsf:
## Overview
This PR resolves **Issue #129**, where the Arrays validation was failing to check rule `20116`. Previously, if an `SBaseRef` pointed to an array (an object with a `ListOfDimensions`) but lacked the required `Index` objects, the validator would silently pass instead of throwing the expected error.

## Changes
* **New Validation Rule**: Created `SBaseRefIndicesCheck.java` in the arrays constraints package. This class safely looks up the referenced element (using `getIdRef()` or `getElementByMetaId()`) and checks if the target has dimensions. If it does, it verifies that the referencing `SBaseRef` contains a `ListOfIndices`.
* **Registered Rule 20116**: Added `ARRAYS_20116` to the `GENERAL_CONSISTENCY` checklist in `ArraysSBasePluginConstraints.java` so the JSBML validation engine actively runs the new check.

## Validation
* Successfully compiled the `arrays` extension (`mvn clean test-compile -pl extensions/arrays`).
* Verified that the changes pass the local test suite (`mvn test -pl extensions/arrays`).

**Closes #129**

### DIFF
--- a/extensions/arrays/src/org/sbml/jsbml/ext/arrays/validator/constraints/SBaseRefIndicesCheck.java
+++ b/extensions/arrays/src/org/sbml/jsbml/ext/arrays/validator/constraints/SBaseRefIndicesCheck.java
@@ -1,0 +1,75 @@
+package org.sbml.jsbml.ext.arrays.validator.constraints;
+
+import java.text.MessageFormat;
+import java.util.ResourceBundle;
+
+import org.sbml.jsbml.Model;
+import org.sbml.jsbml.SBase;
+import org.sbml.jsbml.ext.arrays.ArraysConstants;
+import org.sbml.jsbml.ext.arrays.ArraysSBasePlugin;
+import org.sbml.jsbml.ext.comp.SBaseRef;
+import org.sbml.jsbml.util.ResourceManager;
+import org.sbml.jsbml.validator.offline.factory.SBMLErrorCodes;
+
+/**
+ * Checks that if an SBaseRef points to an array (an object with dimensions),
+ * it must have index objects to dereference it to a scalar.
+ */
+public class SBaseRefIndicesCheck extends ArraysConstraint {
+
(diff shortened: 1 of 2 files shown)


extensions/arrays/src/org/sbml/jsbml/ext/arrays/validator/constraints/SBaseRefIndicesCheck.java
```diff
@@ -0,0 +1,75 @@
...
+import org.sbml.jsbml.ext.arrays.ArraysConstants;
+import org.sbml.jsbml.ext.arrays.ArraysSBasePlugin;
+import org.sbml.jsbml.ext.comp.SBaseRef;
+import org.sbml.jsbml.util.ResourceManager;
+import org.sbml.jsbml.validator.offline.factory.SBMLErrorCodes;
+
+/**
+ * Checks that if an SBaseRef points to an array (an object with dimensions),
+ * it must have index objects to dereference it to a scalar.
+ */
```
draeger, line 17: You may add your `@author` tag.

draeger (approved):

extensions/arrays/src/org/sbml/jsbml/ext/arrays/validator/constraints/SBaseRefIndicesCheck.java
dyrpsf: Done

dyrpsf:
Thanks so much for the review and the merge, @draeger ! 

I really appreciate the reminder to add the tag. I've created a quick follow-up PR to ensure my `@author` tag is attached to the class here: https://github.com/sbmlteam/jsbml/pull/303

Thanks again!

PR #293, opened 2026-04-07
Fix #99: Implement getTypeCode() for compat SBase inheritors

### LINKED ISSUE #99: implementation of the getTypeCode() methods
http://sbml.org/Software/libSBML/docs/java-api/org/sbml/libsbml/SBase.html\#getTypeCode\(\)

public int getTypeCode()
Returns the libSBML type code for this object. 

This method MAY return the typecode of this SBML object or it MAY return SBML_UNKNOWN. That is, subclasses of SBase are not required to implement this method to return a typecode. This method is meant primarily for the LibSBML C interface where class and subclass information is not readily available.

This function might be useful and we would have to implement it anyway for the libsbml-compat module. Would have to be expendable to work with level 3 packages.

Reported by: niko-rodrigue

dyrpsf:
## Overview
This PR resolves **Issue #99** by implementing the `getTypeCode()` method for all `SBase` inheritors within the `libsbml-compat` module, returning the appropriate integer constants to mirror the C++ libSBML API.

## Changes Made
* Created the missing `libsbmlConstants.java` interface to house the standard `SBML_*` integer type codes.
* Added the `getTypeCode()` method to `Species`, `Compartment`, `Model`, `SBMLDocument`, and `ListOfSpecies` within the compatibility module.
* Fixed a pre-existing broken import (`org.sbml.jsbml.xml.test.SBML_L2V1Test`) in `SBMLDocument.java` that was preventing the module from compiling.
* Fixed a pre-existing missing import for `MessageFormat` in `ListOfSpecies.java`.
* Fixed pre-existing missing `throws XMLStreamException` declarations on the `Notes` wrapper methods in `Compartment.java`.

## Validation
* Successfully compiled the `libsbml-compat` module.
* Passed all module tests.

**Closes #99**

### DIFF
--- a/modules/libSBMLcompat/src/org/sbml/libsbml/Compartment.java
+++ b/modules/libSBMLcompat/src/org/sbml/libsbml/Compartment.java
@@ -81,7 +81,7 @@ public class Compartment {
 	/* (non-Javadoc)
 	 * @see org.sbml.jsbml.AbstractSBase#appendNotes(java.lang.String)
 	 */
-	public void appendNotes(String notes) {
+	public void appendNotes(String notes) throws javax.xml.stream.XMLStreamException {
 		compartment.appendNotes(notes);
 	}
 
@@ -123,7 +123,7 @@ public class Compartment {
 	/* (non-Javadoc)
 	 * @see org.sbml.jsbml.AbstractSBase#getNotesString()
 	 */
-	public String getNotesString() {
+	public String getNotesString() throws javax.xml.stream.XMLStreamException {
 		return compartment.getNotesString();
 	}
 
@@ -215,7 +215,7 @@ public class Compartment {
 	/* (non-Javadoc)
 	 * @see org.sbml.jsbml.AbstractSBase#setNotes(java.lang.String)
 	 */
-	public void setNotes(String notes) {
+	public void setNotes(String notes) throws javax.xml.stream.XMLStreamException {
 		compartment.setNotes(notes);
 	}
 
@@ -259,6 +259,14 @@ public class Compartment {
 	 */
 	public void unsetSBOTerm() {
 		compartment.unsetSBOTerm();
+	}
+
+	/**
+	 * Returns the libSBML type code for this object.
+	 * * @return the libSBML type code for this object.
+	 */
+	public int getTypeCode() {
+		return libsbmlConstants.SBML_COMPARTMENT;
 	}
 
 }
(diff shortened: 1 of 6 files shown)


modules/libSBMLcompat/src/org/sbml/libsbml/Compartment.java
```diff
@@ -123,7 +123,7 @@ public Model getModel() {
 	/* (non-Javadoc)
 	 * @see org.sbml.jsbml.AbstractSBase#getNotesString()
 	 */
-	public String getNotesString() {
+	public String getNotesString() throws javax.xml.stream.XMLStreamException {
 		return compartment.getNotesString();
 	}
```
draeger: Why would getting the notes throw an `XMLStreamException`? Setting them activates the parser that may throw an exception, but once it is attached to the object, retrieving the notes shouldn't throw another exeption because the notes had been parsed (and checked) before and are now serialized as `String`.

draeger (changes requested):
I only posted a question. A convincing answer may directly lead to merge.

modules/libSBMLcompat/src/org/sbml/libsbml/Compartment.java
dyrpsf: That is a great question, and you are completely right. 

The reason it was initially throwing an exception is because JSBML's internal `compartment.getNotesString()` dynamically serializes the `XMLNode` tree using a StAX `XMLStreamWriter`, which natively throws a checked `XMLStreamException`. 

However, you are entirely correct that propagating that exception breaks the `libSBMLcompat` contract, because the original C++ libSBML API for `getNotesString()` does not expose any checked exceptions. Since the notes have already been parsed and validated upon insertion, the risk of a serialization failure here is basically zero.

I have just pushed a commit that removes the `throws` clause from the signature and wraps the JSBML call in a `try/catch` block (returning an empty string if it fails) so that it perfectly matches the original C++ API contract again!

dyrpsf:
**Update:** I have pushed a new commit to refine the API compatibility based on the latest review feedback.

* **Strict C++ Compatibility:** Removed the `throws XMLStreamException` clause from `Compartment.getNotesString()`. 
* **Exception Handling:** Wrapped the internal JSBML serialization call in a `try/catch` block. If the XML stream fails, it now gracefully returns an empty string rather than forcing a checked exception onto the caller, perfectly mirroring the original C++ libSBML behavior.

The `libSBMLcompat` test suite passes cleanly locally. Let me know if everything looks good to merge!

draeger (approved):

dyrpsf:
Thanks for putting this together! If there’s anything else you’d like me to tweak or update, just let me know.

PR #295, opened 2026-04-10
Fix #120: Implement canonicalize() method in ASTNode

### LINKED ISSUE #120: Implement canonicalize() method in ASTNode
in our research group we recently switched from using the libsbml 
library to the jsbml library for our program due to performance issues. 
In our old version we had an AST Node parser, that took a AST Node tree 
and converted it into our internal node format. Therefore we used the 
libsbml function "canonicalize()" that checked wether an AST Node of the 
type "Name" is already a predefined function and if so it returned this 
new AST Node. Jsbml is missing such a function.
So my question is, if the library does this implicitly and doesn't 
return name nodes that could be other nodes or if there are any other 
ways to achieve the same result as with the libsbml version.

Reported by: *anonymous

dyrpsf:
## Overview
This PR resolves **Issue #120** by implementing the requested `canonicalize()` method in the `ASTNode` class. This allows the tree to recursively check generic `NAME` or `FUNCTION` nodes and automatically convert them into their specific MathML types (e.g., `CONSTANT_PI`, `FUNCTION_SIN`) if they match a predefined construct.

## Changes Made
* Added `canonicalize()` to `ASTNode.java`. It recursively traverses the tree and uses `Type.getTypeFor(getName())` to safely identify and update node types.
* Updated `TestASTNode.java`. The original C++ ported tests for `canonicalize` were stubbed out as unsupported. I replaced the stub with a working JUnit test (`test_ASTNode_canonicalize`) that successfully verifies constant conversion, function conversion, and recursive traversal.

## Validation
* Successfully compiled `jsbml-core`.
* Verified `test_ASTNode_canonicalize` passes cleanly.

**Resolves #120**

_Note: The commit `b61aade5ae3be0d1a77cb5dc1db570dc23a15e5f` was previously missed due to my mistake and has now been added._

### DIFF
--- a/core/src/org/sbml/jsbml/ASTNode.java
+++ b/core/src/org/sbml/jsbml/ASTNode.java
@@ -5032,4 +5032,28 @@ public class ASTNode extends AbstractTreeNode {
     return plugins;
   }
 
+  /**
+	 * Converts generic NAME or FUNCTION nodes to specific MathML function or 
+	 * constant nodes if their name matches a predefined MathML construct.
+	 * This method recursively traverses the entire AST tree.
+	 */
+	public void canonicalize() {
+		// Check if this node is a generic Name or Function with a valid string
+		if ((getType() == Type.NAME || getType() == Type.FUNCTION) && getName() != null) {
+			
+			// See if the name matches a known MathML constant or function
+			Type specificType = Type.getTypeFor(getName());
+			
+			// If it's a known type, update this node!
+			if (specificType != Type.UNKNOWN && specificType != getType()) {
+				setType(specificType);
+			}
+		}
+		
+		// Recursively canonicalize all child nodes in the tree
+		for (ASTNode child : getChildren()) {
+			child.canonicalize();
+		}
+	}
+
 }
--- a/core/src/org/sbml/jsbml/ASTNode.java
+++ b/core/src/org/sbml/jsbml/ASTNode.java
@@ -5046,7 +5046,7 @@ public class ASTNode extends AbstractTreeNode {
 			
 			// If it's a known type, update this node!
 			if (specificType != Type.UNKNOWN && specificType != getType()) {
-				setType(specificType);
+				setType(getName()); // Pass the string to ensure values like Double.NaN are initialized
 			}
 		}
 		
--- a/core/test/org/sbml/jsbml/test/sbml/TestASTNode.java
+++ b/core/test/org/sbml/jsbml/test/sbml/TestASTNode.java
@@ -109,33 +109,36 @@ public class TestASTNode {
     node = null;
   }
    */
-  /*
-  // Not supported in JSBML (canonicalize method on ASTNode)
-  public void test_ASTNode_canonicalizeConstants()
-  {
-  ...
-	}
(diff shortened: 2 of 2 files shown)


core/src/org/sbml/jsbml/ASTNode.java
```diff
@@ -5032,4 +5032,28 @@ public Map<String, ASTNodePlugin> getExtensionPackages() {
...
+		// Check if this node is a generic Name or Function with a valid string
+		if ((getType() == Type.NAME || getType() == Type.FUNCTION) && getName() != null) {
+			
+			// See if the name matches a known MathML constant or function
+			Type specificType = Type.getTypeFor(getName());
+			
+			// If it's a known type, update this node!
+			if (specificType != Type.UNKNOWN && specificType != getType()) {
+				setType(specificType);
+			}
```
draeger, line 5050: We could always call `setType` because the identical checks are performed within that method. There I also just noticed that there are two specific `String` literals checked in addition to what `getType(<String>)` returns. I wonder if those literals should become part of the checks `getType` performs?

draeger (changes requested):
Little change that may be applied or discarded: Do we need that one check that duplicates another check performed anyway. It may safe very little run time if the check returns false, but if the check returns true, the same check will be performed again.

core/src/org/sbml/jsbml/ASTNode.java
dyrpsf: Thanks for the review, Andreas! I looked into both suggestions and just pushed an update:

**1. The Redundant Check:**
I kept the `if (specificType != Type.UNKNOWN)` check because removing it actually creates a bug. If a user has a variable named `"x"`, `Type.getTypeFor("x")` returns `UNKNOWN`. Calling `setType("x")` unconditionally would overwrite and destroy the variable. 
However, I realized I should be passing the String instead of the Enum. I've updated it to `setType(getName())`, which now properly initializes the numerical values for constants.

**2. The String Literals:**
`Type.getTypeFor(String)` actually *does* catch `"notanumber"` and `"infinity"`, but it returns the generic `Type.REAL` enum for both. Because the Enum loses the specific constant's identity, `setType(String)` still needs those literal checks to manually assign `Double.NaN` and `Double.POSITIVE_INFINITY`. 

Let me know what you think of this approach!

dyrpsf:
**Update:** I have pushed a new commit to refine the `canonicalize()` method based on recent review feedback.

* **Fixed Initialization:** Changed `setType(specificType)` to `setType(getName())` to ensure proper mathematical initialization (e.g., properly assigning `Double.NaN` or `Double.POSITIVE_INFINITY` for specific MathML constants).
* **Variable Protection:** Verified that the `UNKNOWN` type check is strictly necessary to prevent `canonicalize()` from overwriting and destroying user-defined variables.

The `jsbml-core` test suite passes cleanly locally. Ready for further review!

draeger (approved):
Excellent work!

dyrpsf:
> Excellent work!

Thanks for the support and for merging this. Happy to help with any next steps or updates.

PR #296, opened 2026-04-11
Fix #133: Recursively resolve array dimensions in parent objects

### LINKED ISSUE #133: Arrays validation should check for dimensions of parent objects
If an Event has dimension Id 'i', then the child event assignment should be allowed to use 'i' in its math. However, the arrays validation is complaining that 'i' is not a dimension id.

dyrpsf:
## Overview
This PR resolves **Issue #133** where array validation incorrectly threw a "not a dimension id" error when a child object (e.g., an `EventAssignment`) referenced a dimension defined on its parent (e.g., an `Event`).

## Changes Made
* Updated `getDimension(String fieldId)` in `ArraysSBasePlugin.java`. 
* Previously, the method only checked the current object's `ListOfDimensions`, returning `null` if the dimension wasn't defined locally.
* The method now includes a fallback loop that traverses up the SBML tree (`getParentSBMLObject()`), checking the `ArraysSBasePlugin` of each parent to recursively resolve the dimension ID.

## Validation
* Successfully compiled the `jsbml-arrays` extension package.

**Resolves #133**

### DIFF
--- a/extensions/arrays/src/org/sbml/jsbml/ext/arrays/ArraysSBasePlugin.java
+++ b/extensions/arrays/src/org/sbml/jsbml/ext/arrays/ArraysSBasePlugin.java
@@ -544,14 +544,39 @@ public class ArraysSBasePlugin extends AbstractSBasePlugin implements IdManager{
 
   /**
    * Gets an element from the listOfDimensions, with the given id.
+   * If the dimension is not found on this object, it recursively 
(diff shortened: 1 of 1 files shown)


draeger (approved):
Thanks for fixing this issue!

dyrpsf:
> Thanks for fixing this issue!

PR #298, opened 2026-04-14
Fix #216: Enforce strict namespace URI checking for plugins in AbstractSBase

### LINKED ISSUE #216: The test if a plugin is loaded returns true when JSBML has a parser loaded
The test `isSetPlugin` on `AbstractSBase` should return `true` if the `SBase` has a child with the given namespace. In the current implementation, however, it returns `true` if JSBML has (at some point) loaded a parser for parsing that namespace.

For example, if a model has an FBC extension in package version 1, JSBML will return `true` when asking `model.isSetPlugin` with the namespace of FBC package version 2 as argument. For testing, the model 01190 from the SBML test suite can be used.

The desired behavior of this method (and the corresponding method `hasExtension`) is that it should only return `true` if an extension with the given namespace exists.

dyrpsf:
## Overview
This PR resolves **Issue #216** by enforcing strict namespace URI checking when querying for loaded plugins on an `SBase` object.

## Changes Made
Previously, methods like `isSetPlugin()` and `getExtension()` relied purely on the `PackageParser` matching the short label (e.g., `fbc`). This caused a bug where querying a model equipped with an FBC V1 plugin using an FBC V2 namespace URI would incorrectly return `true`, simply because an `fbc` plugin was present in the map.

I updated the following methods in `AbstractSBase.java` to explicitly check that if a full URI is provided, it matches the actual `ElementNamespace` of the loaded plugin:
* `isSetPlugin(String nameOrUri)`
* `getExtension(String nameOrUri)`
* `getPlugin(String nameOrUri)`
* `unsetExtension(String nameOrUri)`

## Validation
* Compiled successfully.
* All 386 `jsbml-core` tests passed cleanly, confirming that the strict matching logic works without breaking existing test cases.

**Resolves #216**

### DIFF
--- a/core/src/org/sbml/jsbml/AbstractSBase.java
+++ b/core/src/org/sbml/jsbml/AbstractSBase.java
@@ -1496,7 +1496,11 @@ public abstract class AbstractSBase extends AbstractTreeNode implements SBase {
         ParserManager.getManager().getPackageParser(nameOrUri);
 
     if (packageParser != null) {
-      return extensions.get(packageParser.getPackageName());
+      SBasePlugin plugin = extensions.get(packageParser.getPackageName());
+      if (plugin != null && !nameOrUri.equals(packageParser.getPackageName()) && !nameOrUri.equals(plugin.getElementNamespace())) {
+        return null;
+      }
+      return plugin;
     }
 
     throw new IllegalArgumentException(format(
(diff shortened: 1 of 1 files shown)


draeger (approved):

dyrpsf:
Thank you for merging this change. I’m happy to iterate further if needed.

PR #299, opened 2026-04-16
feat: Introduce LLMSerializer utility for token-efficient Markdown export

dyrpsf:
## Overview
This PR introduces a new utility class, `LLMSerializer`, designed to bridge JSBML with Large Language Models. It serializes complex SBML `Model` objects into a clean, flat, and token-efficient Markdown format.

## Motivation
Raw SBML XML is highly token-heavy and difficult for LLMs to parse effectively. As part of the groundwork for integrating LLM-based tools with systems biology workflows, this utility extracts only the core biological and mathematical logic (Compartments, Species, and Reactions) and formats it in a highly readable way for both humans and AI.

## Changes Made
* **`LLMSerializer.java`**: Added to `org.sbml.jsbml.util`. It traverses the model and extracts:
  * Compartments and their sizes.
  * Species with initial concentrations/amounts and their compartment locations.
  * Reactions, mapping reactants to products.
  * Kinetic Laws, utilizing `JSBML.formulaToString(ASTNode)` to safely convert the mathematical syntax tree into a flat string equation.
* **`LLMSerializerTest.java`**: Added a comprehensive JUnit test that mocks an SBML Level 3 Version 1 model (with a simple $A \rightarrow B$ reaction) to validate the Markdown serialization.

## Validation
* Compiled successfully without warnings.
* Added `LLMSerializerTest` passes cleanly.

### DIFF
--- a/core/src/org/sbml/jsbml/util/LLMSerializer.java
+++ b/core/src/org/sbml/jsbml/util/LLMSerializer.java
@@ -1,0 +1,96 @@
+package org.sbml.jsbml.util;
+
+import org.sbml.jsbml.Model;
+import org.sbml.jsbml.Species;
+import org.sbml.jsbml.Compartment;
+import org.sbml.jsbml.Parameter;
+import org.sbml.jsbml.Reaction;
+import org.sbml.jsbml.SpeciesReference;
+import org.sbml.jsbml.KineticLaw;
+import org.sbml.jsbml.JSBML;
+
+/**
+ * Utility class to serialize complex SBML models into a flat, 
+ * token-efficient Markdown format optimized for Large Language Models (LLMs).
+ */
+public class LLMSerializer {
+
+    /**
+     * Converts an SBML Model into an LLM-friendly Markdown string.
+     * * @param model The SBML Model to serialize.
+     * @return A Markdown-formatted string representation of the model.
+     */
+    public static String toMarkdown(Model model) {
+        if (model == null) return "Error: Model is null.";
+
+        StringBuilder md = new StringBuilder();
+        
+        // Model Header
+        String modelName = model.isSetName() ? model.getName() : model.getId();
+        md.append("# Model: ").append(modelName).append("\n\n");
+
+        // Compartments
+        md.append("## Compartments\n");
+        for (Compartment c : model.getListOfCompartments()) {
+            String name = c.isSetName() ? c.getName() : c.getId();
+            md.append("* **").append(name).append("** (Size: ").append(c.getSize()).append(")\n");
+        }
+        md.append("\n");
+
+        // Species (The biological entities)
+        md.append("## Species\n");
+        for (Species s : model.getListOfSpecies()) {
+            String name = s.isSetName() ? s.getName() : s.getId();
+            double initialAmt = s.isSetInitialAmount() ? s.getInitialAmount() : s.getInitialConcentration();
+            md.append("* **").append(name).append("**")
+              .append(" (Initial: ").append(initialAmt)
+              .append(", Compartment: ").append(s.getCompartment()).append(")\n");
+        }
+        md.append("\n");
+
+        // Reactions & Math (The core logic)
+        md.append("## Reactions\n");
+        for (Reaction r : model.getListOfReactions()) {
+            String name = r.isSetName() ? r.getName() : r.getId();
+            md.append("* **").append(name).append("**: ");
+            
+            // Build Reactants -> Products string
+            md.append(buildReactionEquation(r)).append("\n");
+
+            // Extract the Math using JSBML's native formula parser!
+            KineticLaw kl = r.getKineticLaw();
+            if (kl != null && kl.isSetMath()) {
+                // This is the magic line that converts the ASTNode tree into a readable math string
+                String mathFormula = JSBML.formulaToString(kl.getMath());
+                md.append("  * Rate Law: `").append(mathFormula).append("`\n");
+            }
+        }
+
+        return md.toString();
+    }
+
+    /**
+     * Helper method to format Reactants -> Products
+     */
+    private static String buildReactionEquation(Reaction r) {
+        StringBuilder eq = new StringBuilder();
+        
+        // Reactants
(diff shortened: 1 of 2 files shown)


core/src/org/sbml/jsbml/util/LLMSerializer.java
```diff
@@ -0,0 +1,96 @@
...
+import org.sbml.jsbml.Parameter;
+import org.sbml.jsbml.Reaction;
+import org.sbml.jsbml.SpeciesReference;
+import org.sbml.jsbml.KineticLaw;
+import org.sbml.jsbml.JSBML;
+
+/**
+ * Utility class to serialize complex SBML models into a flat, 
+ * token-efficient Markdown format optimized for Large Language Models (LLMs).
+ */
```
draeger, line 17: This is a general remark: You may add an `@author` tag where you contributed a class or if you modified it.

core/src/org/sbml/jsbml/util/LLMSerializer.java
```diff
@@ -0,0 +1,96 @@
...
+     * Helper method to format Reactants -> Products
+     */
+    private static String buildReactionEquation(Reaction r) {
+        StringBuilder eq = new StringBuilder();
+        
+        // Reactants
+        for (int i = 0; i < r.getReactantCount(); i++) {
+            SpeciesReference sr = r.getReactant(i);
+            eq.append(sr.getSpecies());
+            if (i < r.getReactantCount() - 1) eq.append(" + ");
```
draeger, line 85: Reactants come with a stoichiometric coefficient that needs to be placed before the reactant (unless the coefficient equals 1.0). Also, if this is to be presented to a human reader, the question is if the name of the referenced `Species` should be displayed rather than its (possibly counter-intuitive) ID.

core/src/org/sbml/jsbml/util/LLMSerializer.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+        
+        eq.append(" -> ");
```
draeger: If the reaction is reversible, a double arrow with harpoon ending should be used: ⇌. For irreversible reactions, we may use the UTF-8 symbol ⟶ that is closer to Chemical standards than `->`.

core/src/org/sbml/jsbml/util/LLMSerializer.java
```diff
@@ -0,0 +1,96 @@
...
+        
+        // Products
+        for (int i = 0; i < r.getProductCount(); i++) {
+            SpeciesReference sr = r.getProduct(i);
+            eq.append(sr.getSpecies());
+            if (i < r.getProductCount() - 1) eq.append(" + ");
```
draeger, line 110: See my comment above regarding reactants. The same applies to products.

core/src/org/sbml/jsbml/util/LLMSerializer.java
draeger: If the reaction has modifiers, those could be added in square brackets after the arrow (unless there is an elegant way in Markdown to write them over the arrow).

draeger (changes requested):
Is this inteded to be an alternative representation for machines only or should human readers see those reports?

core/src/org/sbml/jsbml/util/LLMSerializer.java
dyrpsf: Great suggestion, thank you! I have added the `@author` tag to the class header.

core/src/org/sbml/jsbml/util/LLMSerializer.java
dyrpsf: Done. I've updated the logic to actively query the `Model` for the human-readable `Species` name instead of just printing the ID. I also added a check to print the stoichiometric coefficient gracefully (only if it does not equal 1.0).

core/src/org/sbml/jsbml/util/LLMSerializer.java
dyrpsf: Implemented! I swapped the text arrow for the proper UTF-8 symbols (`⇌` and `→`) based on `r.isReversible()`. I also added a block to append modifiers in square brackets `[...]` immediately following the arrow.

core/src/org/sbml/jsbml/util/LLMSerializer.java
dyrpsf: Applied here as well. I extracted the formatting logic into a shared `appendSpeciesReference()` helper method so both reactants and products get the exact same stoichiometry and name resolution.

dyrpsf:
> Is this inteded to be an alternative representation for machines only or should human readers see those reports?

It is actually meant for both! While the flat structure is optimized to save tokens in LLM context windows, human researchers will definitely need to read these summaries to verify the data being fed to the AI. Because of that, I completely agree we should prioritize human readability and adhere to standard chemical notation.

dyrpsf:
**Update:** I have pushed a new commit implementing all the formatting and notation standards suggested in the recent review. 

* **Chemical Notation:** Swapped text arrows for proper UTF-8 symbols (`⇌`, `→`) and added logic to append modifiers in square brackets.
* **Human Readability:** The serializer now resolves the actual `Species` name instead of printing the raw ID.
* **Stoichiometry:** Reactants and products now correctly display their stoichiometric coefficients if they do not equal 1.0.
* **Documentation:** Added the `@author` tag.

The `core` module compiles and tests cleanly locally. Ready for another look!

draeger (approved):
Only two minor comments:
1. I think there is one unnecessary asterisk before the `@author` tag.
2. There is a small code duplication because reactants and products are both represented with `SpeciesReference`, so it would be possible to shift the listing of those to a private method but this is both optional.

dyrpsf:
Thanks so much for the review and the merge, @draeger !

Even though this was merged, I didn't want to leave that minor technical debt behind. I've created a quick follow-up PR that removes the extra asterisk from the Javadoc and extracts the Reactant/Product loops into a shared private helper method to eliminate the code duplication: PR https://github.com/sbmlteam/jsbml/pull/302

Thanks again for all the help!

PR #301, opened 2026-04-19
Fix #249: Correct unary operator precedence in ASTNode FormulaParser

### LINKED ISSUE #249: Unary operator precedence ignored by ASTNode parser.
Given this expression  (konm * (h ^  - 1.0) / koffm), parsing it with the ASTNode parser will result in an error for the unary operator.
The following equivalent expression will parse correctly:   (konm * (h ^  (0.0 - 1.0)) / koffm)

dyrpsf:
## Overview
This PR resolves **Issue #249** by correcting the operator precedence for unary minus (`-`) and logical NOT (`not` / `!`) in the Level 1/2 `FormulaParser.jj`.

## Changes Made
Previously, the rules for `<MINUS>` and `<NOT>` were located at the very bottom of the parsing hierarchy inside the `Primary()` method. Because of this, the parser gave unary operators the highest possible precedence, incorrectly parsing expressions like `-x^y` as `(-x)^y`. This also caused the parser to throw a `ParseException` when encountering standard parenthetical expressions involving unary operators, such as `(-1.0)`.

I updated `FormulaParser.jj` by:
* Removing the unary operators from `Primary()`.
* Introducing a new `UnaryTerm()` parsing level directly below `TermLvl2()` (Multiplication/Division) and above `TermLvl3()` (Exponentiation).
* Regenerating the parser via JavaCC.

## Validation
* `mvn clean compile` runs cleanly, and the JavaCC parser generates successfully.
* Tested against expressions like `(konm * (-1.0) / koffm)` to ensure the `ParseException` is no longer thrown.

**Resolves #249**

### DIFF
--- a/core/src/org/sbml/jsbml/math/parser/FormulaParser.jj
+++ b/core/src/org/sbml/jsbml/math/parser/FormulaParser.jj
@@ -351,9 +351,9 @@ private ASTNode TermLvl2() :
   ASTNode node = null;
 }
 {
-  leftChild = TermLvl3()
+  leftChild = UnaryTerm()
   (
-    < DIVIDE > rightChild = TermLvl3()
+    < DIVIDE > rightChild = UnaryTerm()
     {
       Integer left, right;
       left = getInteger(leftChild);
@@ -372,7 +372,7 @@ private ASTNode TermLvl2() :
         leftChild = node;
       }
     }
-  | < TIMES > rightChild = TermLvl3()
+  | < TIMES > rightChild = UnaryTerm()
     {
       node = new ASTNode('*');
       node.addChild(leftChild);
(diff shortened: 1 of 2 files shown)


draeger (changes requested):
Do we already have a JUnit test? If not, I'd like to ask for the creation of one.

draeger:
Excellent job, @dyrpsf!

dyrpsf:
> Do we already have a JUnit test? If not, I'd like to ask for the creation of one.

Thanks so much for the review and the bug assignment, @draeger! 

There actually wasn't an automated JUnit test for the Formula Parser in the `core` module yet, so I went ahead and created `core/test/org/sbml/jsbml/text/parser/FormulaParserTest.java` to lock this in.

I've just pushed the new test suite, which specifically targets the two main failure points of Issue #249:
1. **Precedence:** It verifies that `-x^y` correctly evaluates to a root `Type.MINUS` node with a child `Type.POWER` node (meaning `-(x^y)`), ensuring exponentiation correctly takes precedence over the unary minus.
2. **Parentheses:** It parses `(konm * (-1.0) / koffm)` to guarantee that unary operators wrapped in parentheses no longer crash the parser with a `ParseException`.

Let me know if you need anything else!

draeger (approved):

dyrpsf:

PR #306, opened 2026-04-28
fix: Enforce operator precedence brackets in FormulaCompiler (fixes #259)

### LINKED ISSUE #259: Found the bug in ASTNode.toFomula()
Hey, all.
I used jsbml-1.6.1, current latest version of jsbml, but I found the method, ASTNode.toFormula() fail to parse the formula.
To be specific, for testcase, I downloaded BIOMD0000000022.xml from BioModels(https://www.ebi.ac.uk/biomodels/search?query=BIOMD0000000022&domain=biomodels), and I wrote the program in C with libsbml.5.18.0 and java with jsbml-1.6.1.
I attached the both result as libsbml-formula.txt and jsbml-formula.txt, so I wanna you'll check these, then you should compare the first line. 

**libsbml-formula.txt of first line**
> formula1: compartment_0000003 * (c1 + (B1 + pow(CCn / A1, a)) * s1 / (1 + B1 + pow(CCn / A1, a) + pow(PTn / r1, r)))

**jsbml-formula.txt of first line**
> formula1: compartment_0000003*(c1+((B1+CCn/A1^a)*s1)/(1+B1+CCn/A1^a+PTn/r1^r))

They are completely different. And I checked the related report([Robust oscillations within the interlocked feedback model of Drosophila circadian rhythm.](http://identifiers.org/pubmed/11403560), but in this report, result by libsbml is correct.

And, I also checked _core/src/org/sbml/jsbml/ASTNode.java.newASTNode_, this file was revised 2 years ago, but jsbml-1.6.1 was released in Oct 2022. So even in latest version, jsbml has the bug around ASTNode.toFormula().

Any help, please.

dyrpsf:
### Description
This PR resolves Issue #259, which reported incorrect ASTNode string serialization for power functions where the left child possessed a lower mathematical precedence (e.g., division, addition). 

The root cause was located in `FormulaCompiler.java`'s `checkArgumentBrackets` method, which failed to explicitly check for `isOperator()`, `isRelational()`, or `isLogical()` before returning the unbracketed string. 

### Changes Made
* **`FormulaCompiler.java`**: Added an explicit check within `checkArgumentBrackets` to ensure that lower-precedence operators (like `DIVIDE`, `PLUS`, etc.) are wrapped in parentheses when serialized as arguments to higher-precedence operations like `POWER`.
* **`FormulaCompilerLibSBMLTest.java`**: Created a new JUnit test `testIssue259OperatorPrecedence` that perfectly reconstructs the `(CCn/A1)^a` AST tree reported in the original issue to guarantee it now serializes correctly with brackets.

### Testing
Verified that all tests in `jsbml-core` compile and pass cleanly, ensuring this fix does not break any existing mathematical formatting across the library.

### DIFF
--- a/core/src/org/sbml/jsbml/util/compilers/FormulaCompiler.java
+++ b/core/src/org/sbml/jsbml/util/compilers/FormulaCompiler.java
@@ -506,6 +506,12 @@ public class FormulaCompiler extends StringTools implements ASTNodeCompiler {
 
     String term = node.compile(this).toString();
 
+    // Fix for Issue #259: Explicitly enforce brackets for lower-precedence operators 
+    // (DIVIDE, PLUS, MINUS, TIMES) to ensure correct mathematical precedence in power functions.
+    if (node.isOperator() || node.isRelational() || node.isLogical()) {
+      return brackets(term).toString();
+    }
+
     if ((node.isNumber() || node.isString() || node.isFunction()) 
         && (! (node.getType() == ASTNode.Type.FUNCTION_POWER || node.getType() == ASTNode.Type.FUNCTION_REM)))
     {
@@ -513,8 +519,7 @@ public class FormulaCompiler extends StringTools implements ASTNodeCompiler {
     }
 
     // for node.isRelational() and node.isLogical(), we want to put brackets 
-    
-    return term = brackets(term).toString();
+    return brackets(term).toString();
   }
 
   /* (non-Javadoc)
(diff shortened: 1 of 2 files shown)


draeger (approved):
Thanks for this valuable contribution, @dyrpsf!

dyrpsf:
> Thanks for this valuable contribution, @dyrpsf!

PR #307, opened 2026-05-03
Fix: Automatically resolve 301 redirects in comp module tests (#266)

### LINKED ISSUE #266: Module Comp Failing It's Tests
### Problem
The module comp contained within the extensions folder is unable to past all of it's tests because specific external files such as [Spec Example](https://sbmlteam.github.io/jsbml/test/data/comp/spec_example1.xml) return a 301 redirection.

### Recreation
Execute the tests within the comp module.

### Potential Solutions

- Have the input stream which manages these external files automatically resolve 301 redirects
- Change the URL or location of files so that no redirection needs to occur to access them

dyrpsf:
Fixes #266.

### Problem
The `comp` module tests were failing due to `WstxParsingException` errors. This was caused by external resources (such as the `Spec Example` XML) returning `301 Moved Permanently` status codes. Because Java's default `URL.openStream()` does not automatically follow redirects across protocol changes (e.g., HTTP to HTTPS), the parser was attempting to read the HTML redirection page instead of the actual SBML XML file.

### Solution
- Introduced a recursive `openStreamWithRedirects(URL url)` helper method in `ExternalModelDefinition.java`.
- This logic explicitly inspects the `HttpURLConnection` status code and follows the `Location` header for `301`, `302`, `307`, and `308` redirects.
- Replaced the standard `openStream()` call with this new method to ensure robust fetching of external model definitions.

### Testing
Ran `mvn test` in the `extensions/comp` directory. All 20 tests in `TestExternalModelDefinition` and `TestModelDefinitionCloning` now pass successfully with 0 errors.

### DIFF
--- a/extensions/comp/src/org/sbml/jsbml/ext/comp/ExternalModelDefinition.java
+++ b/extensions/comp/src/org/sbml/jsbml/ext/comp/ExternalModelDefinition.java
@@ -19,10 +19,12 @@ package org.sbml.jsbml.ext.comp;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLConnection;
 import java.util.Map;
 
 import javax.xml.stream.XMLStreamException;
@@ -532,7 +534,7 @@ public class ExternalModelDefinition extends AbstractNamedSBase
       logger.info("externalModelDefinition " + getId()
         + " points to an online-resource. Trying to open connection to: "
         + source);
-      InputStream stream = sourceUrl.openStream();
+      InputStream stream = openStreamWithRedirects(sourceUrl);
       externalFile = org.sbml.jsbml.SBMLReader.read(stream);
       externalFile.getSBMLDocument().setLocationURI(sourceUrl.toURI().toString());
       logger.info("Successfully read online source of externalModelDefinition "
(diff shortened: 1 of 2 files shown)


extensions/comp/src/org/sbml/jsbml/ext/comp/ExternalModelDefinition.java
```diff
@@ -682,4 +684,27 @@ public URI getAbsoluteSourceURI(URI absoluteContainingURI) throws MalformedURLEx
...
+
+    if (connection instanceof HttpURLConnection) {
+      HttpURLConnection httpConn = (HttpURLConnection) connection;
+      httpConn.setInstanceFollowRedirects(true); // Enable basic redirect following
+      int status = httpConn.getResponseCode();
+
+      // Manually handle 301/302 redirects if the protocol changes (e.g., HTTP -> HTTPS)
+      if (status == HttpURLConnection.HTTP_MOVED_PERM || 
+        status == HttpURLConnection.HTTP_MOVED_TEMP || 
+        status == 307 || status == 308) {
```
draeger: Could these number literals be defined either as `enum` or `public static final int`, each with a meaningful name, please?

draeger (changes requested):
Minor improvement request: Please avoid using `int` literals and use named constants instead to improve code comprehension.

extensions/comp/src/org/sbml/jsbml/ext/comp/ExternalModelDefinition.java
dyrpsf: Thanks for the suggestion! I have replaced the magic numbers with `HTTP_TEMPORARY_REDIRECT` and `HTTP_PERMANENT_REDIRECT` constants at the top of the class. I also went ahead and added a quick test suite (`ExternalModelDefinitionTest.java`) to verify the constants are correctly defined. Everything is pushed and the `comp` module tests are passing cleanly!

dyrpsf:
**Update:** Just pushed a quick refactor based on the review. 
* Replaced the raw HTTP redirect codes (307/308) with named constants for better readability.
* Added `ExternalModelDefinitionTest.java` to ensure the new constants are covered. 

All module tests are green!

draeger (approved):
Thanks for the hard work, @dyrpsf!

dyrpsf:
> Thanks for the hard work, @dyrpsf!

Thank you for merging this and for the kind words! I’d be glad to make any follow-up changes if something comes up.

PR #308, opened 2026-05-03
Fix: Update initDefaults() visibility to public across all extensions

dyrpsf:
### Problem
The build on `master` was failing across multiple extension modules (`dyn`, `fbc`, `groups`, `layout`, `render`) with the compilation error:
`attempting to assign weaker access privileges; was public`

This occurred because the `initDefaults()` method in the core `SBase` interface was recently updated to `public`, but the inheriting classes in the extension modules were still using weaker access modifiers (package-private or protected), causing a strict Java inheritance conflict.

### Solution
Updated the method signature of `initDefaults()` to `public` in all affected extension files, including:
- `DynElement` & `SpatialComponent` (dyn)
- `GeneProductAssociation`, `ListOfObjectives`, `Objective`, `LogicalOperator` (fbc)
- `Member`, `Group` (groups)
- `Dimensions`, `Curve`, `CurveSegment` (layout)
- `RenderCurveSegment` (render)

### Testing
Ran a full `mvn test` from the root directory and verified a 100% clean build and test pass across all 17 modules.

### DIFF
--- a/extensions/dyn/src/org/sbml/jsbml/ext/dyn/DynElement.java
+++ b/extensions/dyn/src/org/sbml/jsbml/ext/dyn/DynElement.java
@@ -58,7 +58,7 @@ public class DynElement extends AbstractNamedSBase implements UniqueNamedSBase {
   /**
    * Initializes custom Class attributes
    * */
-  private void initDefaults() {
+  public void initDefaults() {
     setPackageVersion(-1);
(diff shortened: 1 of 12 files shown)


draeger (approved):
I agree: This change should be made.

dyrpsf:
> I agree: This change should be made.

PR #309, opened 2026-05-09
fix: prevent infinite recursion in ExpandFunctionDefinitionConverter

dyrpsf:
### Description
This PR addresses the outstanding `TODO` items in `ExpandFunctionDefinitionConverter.java` by introducing a safeguard against infinite recursive loops during macro-expansion, and cleaning up auto-generated stubs.

### Changes Made
1. **Infinite Loop Prevention:** Replaced the unbounded `do-while` loop with a `MAX_ITERATIONS` limit (set to 100). If a malformed or recursive `FunctionDefinition` attempts to expand infinitely, it will now gracefully break and log an error rather than throwing a `StackOverflowError` or hanging the application.
2. **Stub Cleanup:** Removed the auto-generated `// TODO Auto-generated method stub` comment from `setOption` to maintain codebase cleanliness.
3. **Rigorous Testing:** Authored `ExpandFunctionDefinitionConverterTest.java` to verify both standard expansion behavior and the new infinite loop safeguard (utilizing a 2000ms timeout to ensure the application escapes successfully).

All core tests are passing cleanly (`BUILD SUCCESS`). Let me know if you'd prefer a different iteration limit!

### DIFF
--- a/core/src/org/sbml/jsbml/util/converters/ExpandFunctionDefinitionConverter.java
+++ b/core/src/org/sbml/jsbml/util/converters/ExpandFunctionDefinitionConverter.java
@@ -141,7 +141,9 @@ public class ExpandFunctionDefinitionConverter implements SBMLConverter {
     // Starting the actual conversion
     boolean foundFunctionDefinition = false;
     int iterationCount = 0;
-    final int MAX_ITERATIONS = 100; // Safeguard against recursive function definitions
+    // Max depth of valid nested macros is bounded by the total number of functions.
+    // +1 added as a defensive buffer against off-by-one edge cases during traversal.
+    final int MAX_ITERATIONS = m.getFunctionDefinitionCount() + 1; 
 
     // container for the math to be able to modify it inside the Filter if the top level AST is a functionDefinition
     final List<ASTNode> container = new ArrayList<ASTNode>();
--- a/core/src/org/sbml/jsbml/util/converters/ExpandFunctionDefinitionConverter.java
+++ b/core/src/org/sbml/jsbml/util/converters/ExpandFunctionDefinitionConverter.java
@@ -140,11 +140,18 @@ public class ExpandFunctionDefinitionConverter implements SBMLConverter {
 
     // Starting the actual conversion
     boolean foundFunctionDefinition = false;
+    int iterationCount = 0;
+    final int MAX_ITERATIONS = 100; // Safeguard against recursive function definitions
+
     // container for the math to be able to modify it inside the Filter if the top level AST is a functionDefinition
     final List<ASTNode> container = new ArrayList<ASTNode>();
     container.add(math);
 
     do {
+      if (iterationCount++ > MAX_ITERATIONS) {
+        System.err.println("ExpandFunctionDefinitionConverter: Maximum expansion limit reached. Possible infinite recursive loop detected in FunctionDefinitions.");
+        break; // Break out of the loop to prevent application hanging
+      }
 
       List<? extends TreeNode> foundFunctionDefinitions = container.get(0).filter((new Filter() {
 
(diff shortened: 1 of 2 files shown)


core/src/org/sbml/jsbml/util/converters/ExpandFunctionDefinitionConverter.java
```diff
@@ -140,11 +140,18 @@ public static ASTNode expandFunctionDefinition(final Model m, final ASTNode math
 
     // Starting the actual conversion
     boolean foundFunctionDefinition = false;
+    int iterationCount = 0;
+    final int MAX_ITERATIONS = 100; // Safeguard against recursive function definitions
```
draeger: Interesting that we need that. To my knowledge, SBML defines function definitions not as actual mathematical functions but rather as macro definitions. That means, recursion is prohibited. Thus, SBML can't be used as a functional programming language and isn't Turing-complete. By introducing this limit, it seems you identified cases where recursion occurs. Shouldn't this maximum, therefore, be set to one recursive call?

draeger (approved):

core/src/org/sbml/jsbml/util/converters/ExpandFunctionDefinitionConverter.java
dyrpsf: That is a great point regarding the SBML specification prohibiting recursion! 

You are completely right that a valid model should never infinitely recurse.

The main reason I didn't set the maximum strictly to `1` is to account for valid *nested* macro definitions (for example, if `Function A`'s math contains a call to `Function B`). Since the `do-while` loop iterates as long as `foundFunctionDefinition` is true, it might legitimately need a few passes to fully unroll a deeply nested, but perfectly legal, set of distinct macros.

That being said, `100` is definitely overkill given that true recursion is illegal. Would you prefer I lower `MAX_ITERATIONS` to a much tighter limit, like `5` or `10`? This would safely allow for legitimate deep macro nesting while immediately catching and breaking any illegally recursive models!

luciansmith:
Wouldn't the max number of recursions be 'the number of function definitions in the model'?

dyrpsf:
> Wouldn't the max number of recursions be 'the number of function definitions in the model'?

That is a brilliant point, and the logic makes perfect sense! Tying the maximum iterations dynamically to `m.getFunctionDefinitionCount()` is a much more elegant and robust solution than using a hardcoded limit.

Would you like me to go ahead and update the PR to implement this dynamic limit? I am happy to make the change and update the test suite accordingly!

luciansmith:
I am trying and failing to come up with a counter-example that would have more legitimate recursions than there are functions, so I think setting it to that dynamic limit should be fine.  I would personally, purely out of instinct, set the limit to 'nFunctions + 1', even though I think that's probably too large, on the theory that this sort of thing often needs a +1, and a +1 won't hurt if there is indeed recursion; it'll just cycle one extra.

dyrpsf:
> I am trying and failing to come up with a counter-example that would have more legitimate recursions than there are functions, so I think setting it to that dynamic limit should be fine. I would personally, purely out of instinct, set the limit to 'nFunctions + 1', even though I think that's probably too large, on the theory that this sort of thing often needs a +1, and a +1 won't hurt if there is indeed recursion; it'll just cycle one extra.

I completely agree with the "instinctual +1" - it is always better to be defensive and allow one harmless extra cycle just in case of an off-by-one quirk in the AST traversal!

I have updated `MAX_ITERATIONS` to `m.getFunctionDefinitionCount() + 1`. The tests are passing cleanly with the new dynamic limit, and the changes are pushed. Thanks again for the great catch on the math here!

core/src/org/sbml/jsbml/util/converters/ExpandFunctionDefinitionConverter.java
dyrpsf: Just to close the loop on this specific thread - @luciansmith pointed out below that we can dynamically bound this to `m.getFunctionDefinitionCount() + 1`. 

I've implemented that logic and pushed the update, so the hardcoded limit is entirely gone now!

draeger (approved):
Great work! Thanks a lot, @dyrpsf and thanks to @luciansmith for thoughtful comments.

dyrpsf:
> Great work! Thanks a lot, @dyrpsf and thanks to @luciansmith for thoughtful comments.

Thank you for merging this and for the kind words! If there’s anything else you’d like me to tweak or update, just let me know.

PR #305, opened 2026-04-27
feat: Introduce Phase 1 of AntimonySerializer for LLM integration

dyrpsf:
### Description
Following up on discussions regarding the token-efficiency of the Antimony scripting language and potential use-cases for bidirectional IDE plugins (e.g., VSCode-Antimony), this PR introduces the Phase 1 scaffolding for a native JSBML-to-Antimony serializer (`AntimonySerializer.java`). 

By bypassing the XML structure entirely, this utility acts as a highly optimized pipeline for feeding SBML data into Large Language Models and text-editor UIs. 

### Current Implementation (Phase 1)
This initial PR sets up the architecture and successfully translates the foundational structural elements into valid Antimony syntax:
* **Modular Extraction:** Serialization logic is broken into standalone methods (`toAntimony(Species)`, `toAntimony(Compartment)`).
* **Generic Router:** Implemented `toAntimony(SBase element)` to act as a dynamic hook for IDE plugins, allowing selective serialization without traversing the entire `Model`. 
* **Strict Species Initialization:** Implements rigorous checking for `hasOnlySubstanceUnits` and `boundaryCondition` flags to dynamically map SBML initial amounts/concentrations into Antimony's native concentration assumptions (dynamically multiplying or dividing by the compartment ID as necessary).

### Future Scope
I have left a placeholder `TODO` in the code for the more complex serialization logic (Reactions, Events, Algebraic/Rate Rules, and FBC constraints). I plan to build out this advanced mapping logic comprehensively during the upcoming `gsoc-sysbio-llm-tools` coding period. 

### Testing
* Added `AntimonySerializerTest.java` to ensure robust coverage of Phase 1 parsing logic, including the `SBase` dynamic router and the concentration math matrix.
* Verified that `jsbml-core` continues to compile cleanly.

Issue #262, opened 2024-08-12
Extend TreeNodeWithChangeSupport with Visitor pattern


Schmoho:
This is a suggestion: It seems to me that the current implementation does not support the idiomatic way to perform tree walks in Java, because the TreeNode interface that is used does not support the Visitor pattern.

That is, doing a tree walk now requires to explicitly handle type dispatch in the walking class, which is kind of painful to do when also accounting for Plugins and Version differences.

I am very uncertain whether my current understanding of JSBML is correct in that regard and also I am kind of shaky on Java patterns, so please feel free to correct me.

Since the TreeNodeWithChangeSupport interface is the central interface which allows to perform generic tree walks on SBMLDocuments, I would suggest to add at least one interfaces`MutatingTreeNodeVisitor` with signatures `public void visit (Species s)` etc.   TreeNodeWithChangeSupport would then have `void accept(MutatingTreeNodeVisitor v)`.

draeger:
Thank you for this interesting suggestion. In my point of view, the `TreeNode` interface that is on the top of JSBML's type hierarchy was placed in the wrong Java package. There should have been a general algorithm or patterns package but instead this important interface is part of the SWING graphical user interface package. Your idea of making the derived interface `TreeNodeWithChangeSupport` more compatible with recent developments seems sound. However, modifications to a top-level generic interface may have many (potentially also unwanted) effects. In addition, an implementation needs to provided in one of the implementing abstract classes (or a default implementation if suitable). Could you elaborate what subsequent changes to the type hierarchy or implementing classes would be needed?

dyrpsf:
Hi @draeger and @Schmoho ,

I would like to work on this implementation. I agree that the Visitor pattern would provide a much more idiomatic and cleaner way to handle tree walks across the SBML structure.

@draeger , regarding your concern about the type hierarchy and potential unwanted effects: I am happy to map out the specific classes that would be affected. My initial thought is to implement the `accept` method within the central `TreeNodeWithChangeSupport` interface and provide a default implementation in a high-level abstract class (like `AbstractSBase`) to minimize the footprint across subclasses.

Before I move forward with a proposal or PR, I have two questions:
1. Would you prefer separate interfaces for read-only and mutating visitors (e.g., `TreeNodeVisitor` and `MutatingTreeNodeVisitor`), or a single unified visitor interface?
2. Are there specific areas in the SWING-related package hierarchy where you are most concerned about side effects?

[...]

PR #310, opened 2026-05-11
feat: implement named stoichiometry and advanced event options for AntimonySerializer

dyrpsf:
### Description
This PR acts as a direct follow-up to the Phase 1-3 implementation merged in #305. It addresses the advanced edge cases and feature requests highlighted by @luciansmith during the initial review, ensuring the serializer can handle complex SBML models cleanly and token-efficiently.

### Key Updates
* **Named Stoichiometry:** Added logic to `toAntimony(Reaction r)` to check for and serialize named stoichiometries (e.g., dynamically mapping to `n S2 => S3` if the `SpeciesReference` has an ID).
* **Advanced Event Options:** Expanded `toAntimony(Event e)` to support complex event configurations:
  * Delays (`delay=...`)
  * Priorities (`priority=...`)
  * Initial Values (`t0=false`)
  * Persistence (`persistent=false`)
* **Token Efficiency:** The serializer strictly respects SBML defaults. Options like `t0` and `persistent` are only appended if explicitly set to `false`, keeping the resulting Antimony string as concise as possible.

### Testing
Expanded `AntimonySerializerTest.java` to cover these new edge cases. All tests are passing cleanly (`BUILD SUCCESS`).

With these advanced features in place, the underlying serialization logic is fully complete and ready to be adapted into the upcoming `TreeWalk` Visitor pattern we are discussing in #262!

### DIFF
--- a/core/src/org/sbml/jsbml/util/AntimonySerializer.java
+++ b/core/src/org/sbml/jsbml/util/AntimonySerializer.java
@@ -182,7 +182,7 @@ public class AntimonySerializer {
             // Check for Named Stoichiometry (e.g., 'n S2')
             if (sr.isSetId()) {
                 ant.append(sr.getId()).append(" ");
-            } else if (sr.isSetStoichiometry() && sr.getStoichiometry() != 1.0) {
+            } else if (sr.isSetStoichiometry() && sr.getStoichiometry() != 1d) {
                 // Formatting to remove trailing zeros for clean output (e.g. 2.0 -> 2)
                 ant.append(sr.getStoichiometry() == (long) sr.getStoichiometry() ? 
                            String.format("%d", (long)sr.getStoichiometry()) : 
@@ -205,7 +205,7 @@ public class AntimonySerializer {
             
             if (sr.isSetId()) {
                 ant.append(sr.getId()).append(" ");
-            } else if (sr.isSetStoichiometry() && sr.getStoichiometry() != 1.0) {
+            } else if (sr.isSetStoichiometry() && sr.getStoichiometry() != 1d) {
                 ant.append(sr.getStoichiometry() == (long) sr.getStoichiometry() ? 
                            String.format("%d", (long)sr.getStoichiometry()) : 
                            String.format("%s", sr.getStoichiometry())).append(" ");
@@ -272,15 +272,15 @@ public class AntimonySerializer {
 
         // Advanced Event Options
         if (e.isSetPriority() && e.getPriority().isSetMath()) {
-            ant.append(", priority=").append(ASTNode.formulaToString(e.getPriority().getMath()));
+            ant.append(", priority = ").append(ASTNode.formulaToString(e.getPriority().getMath()));
         }
         if (e.isSetTrigger()) {
             org.sbml.jsbml.Trigger t = e.getTrigger();
             if (t.isSetInitialValue() && !t.getInitialValue()) {
-                ant.append(", t0=false");
+                ant.append(", t0 = false");
             }
             if (t.isSetPersistent() && !t.getPersistent()) {
-                ant.append(", persistent=false");
+                ant.append(", persistent = false");
             }
         }
         ant.append(": ");
--- a/core/src/org/sbml/jsbml/util/AntimonySerializer.java
+++ b/core/src/org/sbml/jsbml/util/AntimonySerializer.java
@@ -166,7 +166,7 @@ public class AntimonySerializer {
 
     /**
      * Converts an individual SBML Reaction into an Antimony string.
-     * Handles reactants, products, stoichiometry, reversibility, and kinetic laws.
+     * Handles reactants, products, named stoichiometry, reversibility, and kinetic laws.
      */
     public static String toAntimony(Reaction r) {
         if (r == null) return "";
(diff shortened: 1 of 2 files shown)


dyrpsf:
@draeger, @luciansmith - Just pinging you both here since this PR specifically implements the advanced edge cases we discussed at the end of #305! 

Whenever you have a moment to review the logic for the named stoichiometries and event formatting, let me know if it looks good or if any tweaks are needed!

luciansmith:
That's not actually the correct syntax for delays! It's:

at S1-5 after B>3: x = 4

i.e. '[formula] after'

Are you ensuring that the produced Antimony is correct by trying to parse it with libAntimony?  And then ensuring that libAntimony gives you essentially the same SBML when round-tripped?  It might help with things like this.

dyrpsf:
> That's not actually the correct syntax for delays! It's:
> 
> at S1-5 after B>3: x = 4
> 
> i.e. '[formula] after'
> 
> Are you ensuring that the produced Antimony is correct by trying to parse it with libAntimony? And then ensuring that libAntimony gives you essentially the same SBML when round-tripped? It might help with things like this.

Ah, thank you so much for the correction! I completely misunderstood the Antimony documentation regarding the delay syntax. `at [delay] after [trigger]` makes perfect sense. 

Regarding your question about round-trip testing with `libAntimony`: that is a fantastic suggestion. Currently, the JUnit test suite in JSBML is purely doing string/regex matching against expected literals since JSBML is a pure Java environment. However, moving forward, I will definitely use the Python `libantimony` bindings locally to manually validate these complex strings and ensure they parse and round-trip correctly!

I will push a fix for the delay syntax right away.

dyrpsf:
@luciansmith Fix pushed! The tests are passing cleanly with the updated `at [delay] after [trigger]` syntax. Thanks again for catching that and for the tip on using `libAntimony` for round-tripping!

core/src/org/sbml/jsbml/util/AntimonySerializer.java
```diff
@@ -178,7 +178,11 @@ public static String toAntimony(Reaction r) {
         // 2. Reactants
         for (int i = 0; i < r.getReactantCount(); i++) {
             SpeciesReference sr = r.getReactant(i);
-            if (sr.isSetStoichiometry() && sr.getStoichiometry() != 1.0) {
+            
+            // Check for Named Stoichiometry (e.g., 'n S2')
+            if (sr.isSetId()) {
+                ant.append(sr.getId()).append(" ");
+            } else if (sr.isSetStoichiometry() && sr.getStoichiometry() != 1.0) {
```
draeger: Very minor comment but I usually write `1d` rather than `1.0` because the latter could also be interpreted as a float value. Similarly, I'd use `1f` and `1L` to make clear what is meant.

core/src/org/sbml/jsbml/util/AntimonySerializer.java
```diff
@@ -250,11 +258,32 @@ public static String toAntimony(Event e) {
...
 
-        ant.append("at (");
-        if (e.isSetTrigger() && e.getTrigger().isSetMath()) {
+        ant.append("at ");
+        boolean hasTrigger = e.isSetTrigger() && e.getTrigger().isSetMath();
+        boolean hasDelay = e.isSetDelay() && e.getDelay().isSetMath();
+
+        if (hasDelay && hasTrigger) {
+            ant.append(ASTNode.formulaToString(e.getDelay().getMath()));
+            ant.append(" after ");
```
draeger, line 267: Also, it can make sense to collect Antimony keywords as `public static final String` values in some interface that we can then implement to inherit all the constants directly.

core/src/org/sbml/jsbml/util/AntimonySerializer.java
```diff
@@ -250,11 +258,32 @@ public static String toAntimony(Event e) {
...
-        ant.append("): ");
+
+        // Advanced Event Options
+        if (e.isSetPriority() && e.getPriority().isSetMath()) {
+            ant.append(", priority=").append(ASTNode.formulaToString(e.getPriority().getMath()));
+        }
+        if (e.isSetTrigger()) {
+            org.sbml.jsbml.Trigger t = e.getTrigger();
+            if (t.isSetInitialValue() && !t.getInitialValue()) {
+                ant.append(", t0=false");
```
draeger: Can spaces be added in the output, i.e., `t0 = false`?

core/src/org/sbml/jsbml/util/AntimonySerializer.java
```diff
@@ -250,11 +258,32 @@ public static String toAntimony(Event e) {
...
+            }
+            if (t.isSetPersistent() && !t.getPersistent()) {
+                ant.append(", persistent=false");
```
draeger: Here also blanks please.

draeger (changes requested):
A few minor change requests about formatting only.

core/src/org/sbml/jsbml/util/AntimonySerializer.java
dyrpsf: Fixed! Updated to `1d` to ensure it is explicitly treated as a double.

core/src/org/sbml/jsbml/util/AntimonySerializer.java
dyrpsf: Good catch, added the requested spacing here for better readability!

core/src/org/sbml/jsbml/util/AntimonySerializer.java
dyrpsf: Spacing added here too!

draeger (approved):
I'll merge this now and we can add an interface collecting Antimony keywords later.

dyrpsf:
Thank you for the merge, @draeger! 

Funny timing - I actually just finished building that exact `AntimonyConstants` interface locally right as you merged this! Since the code is already fully written and tested, I will go ahead and open a quick, separate PR for it now so we have that clean architecture locked in.

Update: new PR : #311 .

core/src/org/sbml/jsbml/util/AntimonySerializer.java
dyrpsf: Done (PR #311).

PR #311, opened 2026-05-11
refactor: extract Antimony keywords into AntimonyConstants interface

dyrpsf:
### Description
Following up immediately on @draeger's suggestion at the end of #310. 

This PR extracts all Antimony-specific keywords, operators, and formatting strings into a centralized `AntimonyConstants` interface. `AntimonySerializer` now implements this interface, inheriting the constants directly. 

This removes all hardcoded magic strings from the serializer logic, making the parser significantly cleaner and safer to maintain moving forward. 

All core tests (`AntimonySerializerTest.java`) continue to pass cleanly!

### DIFF
--- a/core/src/org/sbml/jsbml/util/AntimonyConstants.java
+++ b/core/src/org/sbml/jsbml/util/AntimonyConstants.java
@@ -1,0 +1,27 @@
+package org.sbml.jsbml.util;
+
+/**
+ * Interface defining constant keywords and operators used in the Antimony scripting language.
+ * Implementing this interface allows direct inheritance of all constants.
+ *
+ * @author Deepak Yadav
+ */
+public interface AntimonyConstants {
+    public static final String MODEL = "model ";
+    public static final String END = "end\n";
+    public static final String COMPARTMENT = "compartment ";
+    public static final String SPECIES = "species ";
+    public static final String SUBSTANCE_ONLY = "substanceOnly species ";
+    public static final String IN = " in ";
+    public static final String IRREVERSIBLE = " => ";
+    public static final String REVERSIBLE = " -> ";
+    public static final String ASSIGNMENT = " := ";
+    public static final String RATE = "' = ";
+    public static final String ALGEBRAIC = "0 = ";
+    public static final String AT = "at ";
+    public static final String AFTER = " after ";
+    public static final String DELAY = ", delay = ";
+    public static final String PRIORITY = ", priority = ";
+    public static final String T0_FALSE = ", t0 = false";
+    public static final String PERSISTENT_FALSE = ", persistent = false";
+}
--- a/core/src/org/sbml/jsbml/util/AntimonyConstants.java
+++ b/core/src/org/sbml/jsbml/util/AntimonyConstants.java
@@ -7,21 +7,21 @@ package org.sbml.jsbml.util;
  * @author Deepak Yadav
  */
 public interface AntimonyConstants {
-    public static final String MODEL = "model ";
-    public static final String END = "end\n";
-    public static final String COMPARTMENT = "compartment ";
-    public static final String SPECIES = "species ";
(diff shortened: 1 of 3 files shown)


Schmoho:
Here too I would propose to address #313

dyrpsf:
I just pushed `AntimonyConstantsTest.java` to ensure the interface variables are properly verified against regressions! Let me know if everything looks good to go.

dyrpsf:
Just pushed a quick follow-up commit to add the proper Javadoc and `@author` tags to the new test file so it perfectly matches the repository's documentation standards!

core/src/org/sbml/jsbml/util/AntimonyConstants.java
```diff
@@ -0,0 +1,27 @@
...
+    public static final String REVERSIBLE = " -> ";
+    public static final String ASSIGNMENT = " := ";
+    public static final String RATE = "' = ";
+    public static final String ALGEBRAIC = "0 = ";
+    public static final String AT = "at ";
+    public static final String AFTER = " after ";
+    public static final String DELAY = ", delay = ";
+    public static final String PRIORITY = ", priority = ";
+    public static final String T0_FALSE = ", t0 = false";
+    public static final String PERSISTENT_FALSE = ", persistent = false";
```
draeger: While I am always in favor of human-readable code and consequently support the use of spaces, I also think those optional style characters should not be part of a keyword selection. Instead, I think it is better to include them where the source code is generated. This might be a bit extra work, but it allows a more flexible way to generate Antimony output, namely, the number of spaces can be regulated. You can see this, for instance, in source-code editors (such as Eclipse or IntelliJ), which offer settings to adjust the character (tab or blank) and their counts in generated code.

core/src/org/sbml/jsbml/util/AntimonySerializer.java
```diff
@@ -62,7 +62,7 @@ public static String toAntimony(Model model) {
         StringBuilder ant = new StringBuilder();
         
         String modelName = model.isSetName() ? model.getName() : model.getId();
-        ant.append("model ").append(modelName).append("()\n\n");
+        ant.append(MODEL).append(modelName).append("()\n\n");
```
draeger: The `String` generation might need a subsequent change if the blank is removed from the keyword representing the constant field.

draeger (changes requested):
We need to change this again because the output formatting should not be part of the keyword collection. Blanks are good and useful, but their number and kind (tab or space) need to be separatable adjustable. For version 1, we don't necessarily need to implement various code styles, but we should remove blank spaces from keywords. Thus, all uses of the keyword constants need to reintroduce blanks if needed.

core/src/org/sbml/jsbml/util/AntimonyConstants.java
dyrpsf: Done

core/src/org/sbml/jsbml/util/AntimonySerializer.java
dyrpsf: Done

dyrpsf:
Hi @draeger, thanks for the review! 

You are completely right about separating the formatting from the core keywords. I have refactored `AntimonyConstants` to strip out all hardcoded blanks, commas, and newlines so they exist purely as the keyword strings. I then updated `AntimonySerializer` to manually append the spaces and formatting during the generation step. 

This should make it much easier to implement configurable code styles (like tabs vs. spaces) in the future. Let me know if everything looks good to go!

draeger (approved):
Excellent, thanks (even though it is now a bit more combersome but better)!

dyrpsf:
> Excellent, thanks (even though it is now a bit more combersome but better)!

Thank you so much for the review and for merging! I completely agree—it is a small trade-off in convenience for the developer, but it makes for a much cleaner and more scalable architecture long-term. Glad to have this piece officially in!

PR #312, opened 2026-05-13
feat: introduce core TreeNodeVisitor pattern scaffolding (#262)

### LINKED ISSUE #262: Extend TreeNodeWithChangeSupport with Visitor pattern

dyrpsf:
### Description
This PR implements the foundational architecture for the `TreeNodeVisitor` pattern discussed in #262. It provides a clean, generic way to traverse the JSBML AST without modifying core classes for every new feature.

### Key Changes
* **`TreeNodeVisitor<T>`:** Created the core interface with `visit(SBase)` as the primary method and `visit(TreeNodeWithChangeSupport)` as the fallback for auxiliary nodes.
* **`TreeNodeWithChangeSupport`:** Added the `<T> T accept(TreeNodeVisitor<T> visitor)` contract to the root interface.
* **Java 7 Compliance:** Since JSBML enforces Java 7 (which does not support `default` interface methods), the `accept` method implementation has been explicitly injected into the 18 required base classes (e.g., `AbstractSBase`, `ASTNode`, `XMLNode`, and the various math nodes).

All core tests are passing cleanly locally. 

This PR serves purely as the structural scaffolding. Once this architecture is approved and merged, we can begin adapting tools like the `AntimonySerializer` to actively utilize this visitor!

### DIFF
--- a/core/src/org/sbml/jsbml/TreeNodeVisitor.java
+++ b/core/src/org/sbml/jsbml/TreeNodeVisitor.java
@@ -1,0 +1,30 @@
+package org.sbml.jsbml;
+
+import org.sbml.jsbml.util.TreeNodeWithChangeSupport;
+/**
+ * A generic visitor interface for traversing the JSBML tree structure.
+ * This allows external classes to operate on SBML components without 
+ * modifying the core classes.
+ *
+ * @param <T> The return type of the visitor operations.
+ * @author Deepak Yadav
+ */
+public interface TreeNodeVisitor<T> {
+
+    /**
+     * Primary traversal method for core SBML components.
+     *
+     * @param sbase the core SBML element to visit
+     * @return a result of type T
+     */
+    T visit(SBase sbase);
+
+    /**
+     * Fallback traversal method for auxiliary nodes (e.g., annotations, notes)
+     * that do not inherit from SBase but do inherit from TreeNodeWithChangeSupport.
+     *
+     * @param node the tree node to visit
+     * @return a result of type T
+     */
+    T visit(TreeNodeWithChangeSupport node);
+}
--- a/core/src/org/sbml/jsbml/TreeNodeVisitor.java
+++ b/core/src/org/sbml/jsbml/TreeNodeVisitor.java
@@ -1,12 +1,14 @@
 package org.sbml.jsbml;
 
 import org.sbml.jsbml.util.TreeNodeWithChangeSupport;
+
 /**
  * A generic visitor interface for traversing the JSBML tree structure.
- * This allows external classes to operate on SBML components without 
+ * This allows external classes to operate on SBML components without
  * modifying the core classes.
  *
- * @param <T> The return type of the visitor operations.
+ * @param <T> The return type of the visitor operations. Note that this generic 
+ * type may also be {@link Void} if nothing is to be returned.
  * @author Deepak Yadav
  */
 public interface TreeNodeVisitor<T> {
--- a/core/src/org/sbml/jsbml/ext/AbstractASTNodePlugin.java
+++ b/core/src/org/sbml/jsbml/ext/AbstractASTNodePlugin.java
@@ -31,6 +31,7 @@ import org.sbml.jsbml.ASTNode;
 import org.sbml.jsbml.AbstractTreeNode;
 import org.sbml.jsbml.SBMLDocument;
 import org.sbml.jsbml.SBase;
+import org.sbml.jsbml.TreeNodeVisitor;
 import org.sbml.jsbml.util.TreeNodeChangeEvent;
 
 /**
@@ -363,7 +364,7 @@ public abstract class AbstractASTNodePlugin extends AbstractTreeNode implements 
      * @return the result of the visitor operation
      */
     @Override
-    public <T> T accept(org.sbml.jsbml.TreeNodeVisitor<T> visitor) {
+    public <T> T accept(TreeNodeVisitor<T> visitor) {
         return visitor.visit(this);
     }
 
--- a/core/src/org/sbml/jsbml/ext/AbstractASTNodePlugin.java
+++ b/core/src/org/sbml/jsbml/ext/AbstractASTNodePlugin.java
@@ -355,4 +355,16 @@ public abstract class AbstractASTNodePlugin extends AbstractTreeNode implements 
     return new TreeMap<String, String>();
   }
 
+  /**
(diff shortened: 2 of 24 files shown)


dyrpsf:
The mock visitor testing is officially in place! I added `TreeNodeVisitorTest.java` using a custom `DummyVisitor` to verify the generic `<T>` scaffolding works perfectly. 

To resolve the cascading interface implementation requirements across the JSBML extensions, I added the standard `accept` method to the foundational `AbstractSBasePlugin` and `AbstractASTNodePlugin` classes (as well as `RelAbsVector`). The entire multi-module project now compiles flawlessly! Let me know if everything looks good to go.

core/src/org/sbml/jsbml/ext/AbstractASTNodePlugin.java
```diff
@@ -355,4 +355,16 @@ public Map<String, String> writeXMLAttributes() {
...
 
+  /**
+     * Accepts a generic visitor to traverse this component.
+     *
+     * @param <T> the return type of the visitor
+     * @param visitor the visitor implementation
+     * @return the result of the visitor operation
+     */
+    @Override
+    public <T> T accept(org.sbml.jsbml.TreeNodeVisitor<T> visitor) {
```
draeger: Question: Why not importing `TreeNodeVisitor` but instead writing the full prefix `org.sbml.jsbml.` for the type?

core/src/org/sbml/jsbml/ext/AbstractSBasePlugin.java
```diff
@@ -327,4 +327,16 @@ public Map<String, String> writeXMLAttributes() {
...
 
+  /**
+     * Accepts a generic visitor to traverse this component.
+     *
+     * @param <T> the return type of the visitor
+     * @param visitor the visitor implementation
+     * @return the result of the visitor operation
+     */
+    @Override
+    public <T> T accept(org.sbml.jsbml.TreeNodeVisitor<T> visitor) {
```
draeger: Same: I'd suggest adding an import statement and removing the package prefix from the argument here.

core/src/org/sbml/jsbml/TreeNodeVisitor.java
```diff
@@ -0,0 +1,30 @@
+package org.sbml.jsbml;
+
+import org.sbml.jsbml.util.TreeNodeWithChangeSupport;
+/**
+ * A generic visitor interface for traversing the JSBML tree structure.
+ * This allows external classes to operate on SBML components without 
+ * modifying the core classes.
+ *
+ * @param <T> The return type of the visitor operations.
```
draeger: Please add the remark that the generic type may also be `Void` if nothing is to be returned. Java programmers should know that but it is better to point this out again.

extensions/render/src/org/sbml/jsbml/ext/render/RelAbsVector.java
```diff
@@ -327,4 +328,17 @@ public int getChildCount() {
...
+
+  /**
+     * Accepts a generic visitor to traverse this component.
+     *
+     * @param <T> the return type of the visitor
+     * @param visitor the visitor implementation
+     * @return the result of the visitor operation
+     */
+    @Override
+    public <T> T accept(org.sbml.jsbml.TreeNodeVisitor<T> visitor) {
```
draeger: Please import the interface rather than using the full package prefix.

draeger (changes requested):
It is a bit surprising to me that the inherited method is to be implemented in so many classes. JSBML has a very detailed and abstract type hierarchy. I am convinced it is possible to exploit this hierarchy better so that the interface method is implemented only in a very few top-level classes and can be removed from most more specific classes; in particular, all the specific ASTNode types should inherit the method from a more abstract class. Please check the users' guide again, which contains a class diagram, see https://github.com/sbmlteam/jsbml/raw/master/doc/user_guide/User_Guide.pdf
Also, I left a few comments suggesting that import statements could simplify the type declarations of arguments passed to a method.

core/src/org/sbml/jsbml/ext/AbstractASTNodePlugin.java
dyrpsf: Done! I've removed the fully qualified prefix and added the standard import at the top.

core/src/org/sbml/jsbml/ext/AbstractSBasePlugin.java
dyrpsf: Fixed this import as well to match the standard format.

core/src/org/sbml/jsbml/TreeNodeVisitor.java
dyrpsf: Good catch! I've updated the Javadoc to explicitly mention that the generic type can be `Void` if nothing is returned.

extensions/render/src/org/sbml/jsbml/ext/render/RelAbsVector.java
dyrpsf: Updated the import here too. Thanks for the review!

dyrpsf:
Hi @draeger , thanks for the review! 

I have implemented all the requested changes:
* Updated the `TreeNodeVisitor` Javadoc to explicitly clarify the `Void` return type usage.
* Cleaned up the method signatures in `RelAbsVector`, `AbstractSBasePlugin`, and `AbstractASTNodePlugin` to use standard imports instead of fully qualified package prefixes.

The multi-module build is completely clean and passing. Let me know if everything looks good to merge!

draeger (approved):
Great job, thanks!

dyrpsf:

PR #321, opened 2026-08-05
feat: implement reflection-based abstraction layer for AntimonySerializer (#320)

### LINKED ISSUE #320: Refactor `AntimonySerializer` to use abstraction and reflection-based design
### Background
During the implementation of `Parameter` and `FunctionDefinition` support in PR #319, @draeger suggested an architectural improvement for the `AntimonySerializer`. Currently, the serializer relies heavily on individual concrete class handling—specifically, an extensive `instanceof` routing chain and manual `isSet...()` checks for individual attributes.

### Proposed Architecture
To reduce boilerplate and improve the scalability of the utility, we should leverage JSBML's existing abstraction mechanisms. By utilizing entry points like `writeXMLAttributes` or Java reflection (similar to JSBML's `toString` implementation), we can collect attribute/value pairs and structural information at a higher abstraction level.

For attributes or keywords that differ between SBML and Antimony, a simple mapping layer (such as a lookup table or properties-based configuration) can be introduced.

### Implementation Goals
- [ ] **Base Abstraction Layer:** Implement a generic generic serialization method that automatically loops through structural information for an `SBase` element.
- [ ] **Keyword Mapping:** Create a lookup table to translate JSBML attributes to Antimony syntax where they differ.
- [ ] **Generic Element Migration:** Migrate "flat" components (e.g., `Compartment`, `Parameter`, basic `Species`) to inherit their Antimony serialization almost entirely from the generic superclass logic.
- [ ] **Custom Overrides:** Identify and isolate complex components (e.g., `Reaction` kinetic laws, `Event` triggers/assignments) that will still require custom override methods for nested `ASTNode` math formatting.

### Benefits
* Reduces the number of class-specific special cases.
* Eliminates boilerplate attribute-checking code.
* Makes the serializer much easier to maintain and extend for future SBML components.

dyrpsf:
### Description
This PR introduces a major architectural improvement to the `AntimonySerializer`, directly addressing the feedback and approach suggested by @draeger in #320. 

To reduce boilerplate and improve scalability, I have implemented a generic, reflection-based abstraction layer for flat SBML components.

### Key Changes
* **Base Abstraction Layer:** Implemented a `serializeFlatComponent()` engine that dynamically loops through zero-argument `isSet...()` methods using Java Reflection, eliminating the need for manual attribute checks.
* **Keyword Mapping & Filtering:** Added a case-insensitive `ATTRIBUTE_MAP` to translate JSBML attributes directly to Antimony syntax. It also acts as a filter to safely ignore internal JSBML metadata (e.g., `version`, `level`, `parent`).
* **Duplicate Prevention:** Integrated a `HashSet` tracking system within the reflection loop to prevent duplicate values from being serialized (successfully handling JSBML aliases where both `isSetSize()` and `isSetVolume()` might trigger).
* **Component Migration:** Successfully migrated `Compartment` and `Parameter` to utilize this new generic superclass logic. Complex nested components (like `Reaction` and `Event`) remain safely isolated with custom overrides.

### Testing
* All existing tests in `AntimonySerializerTest.java` pass cleanly locally. The architectural rewrite produces the exact same, accurate Antimony string output without any regressions.

Closes #320

### DIFF
--- a/core/src/org/sbml/jsbml/util/AntimonySerializer.java
+++ b/core/src/org/sbml/jsbml/util/AntimonySerializer.java
@@ -1,5 +1,9 @@
 package org.sbml.jsbml.util;
 
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+
 import org.sbml.jsbml.SBase;
 import org.sbml.jsbml.Model;
 import org.sbml.jsbml.Species;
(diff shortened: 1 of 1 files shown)


draeger (approved):
Great job!

dyrpsf:
> Great job!

Thanks for the support and for merging this. Happy to help with any next steps or updates.

